PR #46, opened 2021-10-16
add playback speed control to the player

NIK679:
I have added playback speed control feature for the experimental MPV player and the exoplayer.


jarnedemeulemeester:
Looks good!
I only have one small request, could you change the icon to use [this one](https://lucide.dev/icon/gauge?search=gauge) from Lucide. All the icons the app uses are from that project (if available of course). 
And maybe use the filename `ic_gauge` to reflect the icon name on the site.

Thanks!

PR #47, opened 2021-10-17
Add person detail screen

### LINKED ISSUE #40: Basic per Actor view
It would be nice to have a very simple list of movies and series the actor has appeared in upon clicking the actor picture. Maybe in the style of the current Episode Pop-up?

lsrom:
Displays actor/actresses portrait and text info + list of movies/show this person starred in. Text info is max 5 lines with View More button if ellipsized. View More toggle is reset upon orientation change since in landscape mode ellipsize might not be necessary.

Closes #40 

Few notes:
- I never before worked with the BindingAdapters, hope I didn't make complete mess of it.
- I added enum for type Movie/Series. I think it is good idea to remove conditions based on strings from the code base but I didn't want to change all the files in here in my first contribution so I used it only in the new code.
- I didn't put any effort into landscape mode.

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -84,8 +84,8 @@ dependencies {
     implementation("androidx.preference:preference-ktx:$preferenceVersion")
 
     // Jellyfin
-    val jellyfinVersion = "1.0.3"
-    implementation("org.jellyfin.sdk:jellyfin-platform-android:$jellyfinVersion")
+    val jellyfinVersion = "1.1.0"
+    implementation("org.jellyfin.sdk:jellyfin-core:$jellyfinVersion")
 
     // Glide
     val glideVersion = "4.12.0"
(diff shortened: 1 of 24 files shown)


lsrom:
The `StarredInAdapter` was a mistake, I didn't really had an overview of the codebase so I removed it. 

What issues do you have with the UI? I can work in your feedback if you don't want to di it yourself.

Also sorry for coming here out of the blue and wanting to change things and stuff :) I want to use this app as my main TV app because it's codebase is much better than Jellyfin native app and I have a lot of time on my hands now so you'll here from me a lot I think in coming days.

jarnedemeulemeester:
Haha no problem, I do feel a bit overwhelmed but I really appreciate all the work you and others put in. My code is obviously not perfect. This is also my first Android and open source project so I'm still quite new to it all.

As for the UI, I want to make sure it feels like the other parts of the app. If every screen gets designed and created by different people than there is no coherence between them (if not working closely as a team). I will create some new designs on Figma soon for the person screen and maybe even Android TV. Don't worry to much about the design, I have no problem working on it if you don't mind.

Also maybe it's a good idea to create a Discord server (or something else) to more easily communicate about development and progress on the project. What do you think?

PR #51, opened 2021-10-19
Add offline playback

Jcuhfehl:
I have added the feature that allows the user to download media and play it back when offline.


jarnedemeulemeester:
This can be improved a lot in a future merge request. But it's a good basic implementation for now.
Thanks a lot!

PR #68, opened 2021-11-14
Improve offline playback

Jcuhfehl:
This pull request adds the following improvements to the offline playback feature:

- Fix a bug in download playback tracking
- Remove the unused external storage permission
- Add the overview text to downloaded metadata
- Add a visual indicator of whether an item is downloaded or not
- Use the downloaded item if it is available instead of streaming it from the server


jarnedemeulemeester:
Nice!

I should have asked this on the previous MR but why not use a Room database to store the metadata of downloaded items?
Maybe there is a good reason, I'm just wondering.

PR #74, opened 2022-01-08
Player improvement

maulik9898:
- No more false brightness and volume trigger on status bar swipe [ Exoplayer , MPV]
- Remember last brightness set [ Exoplayer , MPV]
- Pinch to zoom and auto rotate Player [ Exoplayer ]

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -21,7 +21,8 @@
 
         <activity
             android:name=".PlayerActivity"
-            android:screenOrientation="userLandscape" />
+            android:configChanges="screenSize|smallestScreenSize|screenLayout|orientation|keyboard|keyboardHidden|uiMode"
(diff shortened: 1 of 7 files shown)


app/src/main/AndroidManifest.xml
```diff
@@ -21,7 +21,8 @@
 
         <activity
             android:name=".PlayerActivity"
-            android:screenOrientation="userLandscape" />
+            android:configChanges="screenSize|smallestScreenSize|screenLayout|orientation|keyboard|keyboardHidden|uiMode"
```
jarnedemeulemeester, line 24: I don't know exactly what this does. I read that if you set values here you have to manually handle configuration changes in the activity. 

Could you explain what exactly this does and why?

maulik9898: I forgot to remove that line. Yes this for manually handle configuration which is not needed.

PR #76, opened 2022-01-13
EpisodeBottomSheetFragment expanded by default

NIK679:
When clicking on an episode, the bottomsheet opens in collapsed state in landscape mode and has to be dragged up before it can be played. I have changed the behaviour to be expanded, so that it is easily usable in landscape mode.


NIK679:
This change removes the rounded corners on the dialog when it's in expanded state, So I'm not sure if this is the best method.
material-components/material-components-android/#1278

jarnedemeulemeester:
Yeah the rounded corners dissapear in expanded state, that's indeed unfortunate but not the biggest problem.
We could try to implement a workaround mentioned in the link you posted. But for now it's okay :)

PR #79, opened 2022-01-18
Fix downloads getting automatically deleted by android

Jcuhfehl:
Fixed a bug where android would automatically delete downloads because it assumed that findroid had forgotten about them.


jarnedemeulemeester:
I don't think this makes much sense. 
That method is deprecated and the default is already True.

Feel free to prove me otherwise :slightly_smiling_face:

Jcuhfehl:
I just tested my PR again, and it turns out that it actually didn't fix the issue (I just hadn't set my date far enough ahead to trigger the deletion), even though this change should actually fix the problem according to this stackoverflow thread: https://stackoverflow.com/questions/46501822/downloaded-files-get-deleted-automatically

Unfortunately, this means the problem still isn't solved.

PR #80, opened 2022-01-18
Display downloaded episodes by series

Jcuhfehl:
With this pull request, downloaded episodes will now be displayed by the series they are in. This makes the downloaded episodes much easier to navigate.


jarnedemeulemeester:
I will take a look at this as soon as possible

PR #85, opened 2022-02-01
Add switch to toggle downloads over mobile network

MichaelArkh:
Added a switch in settings to enable the change.


jarnedemeulemeester:
Nice but I think it would be better to add two options in settings instead of the Wi-Fi only one:
- "Download using mobile data" -> download_mobile_data
- "Download when roaming" -> download_roaming

PR #100, opened 2022-03-01
Add option to turn off subtitles

### LINKED ISSUE #82: Disable subtitles
When I open an mkv that has a subtitle track with default set to true, findroid automatically enables it as expected. But I would like to disable this functionality, and never have subtitles automatically turned on, regardless of that flag. When I click the caption icon, I only have the options "Automatic" and "English" with no "None" option. I couldn't find anything relevant in the settings either.

Theta-Dev:
I added the option to turn off the subtitles for a video which has subtitles enabled by default. Turns out you just have to enable the `ShowDisableOption` in the `TrackSelectionDialogBuilder`.

This fixes #82.

Here is a screenshot of the new option ("Ohne" = None, my phone is set to German).
Screenshot_20220301-163636


jarnedemeulemeester:
Thanks!

clean-interaction:
I don't see this functionality when I enable mpv player and I can't play the file without enabling mpv so please consider adding this to mpv.

PR #107, opened 2022-04-13
Update new strings for spanish

newton181:
Refactored to match with default string file, minor fiexs and new strings translated.


jarnedemeulemeester:
I have just set up [Weblate](https://weblate.jdtech.dev/projects/findroid/android/) for future translations
If you plan to contribute more translations please use that :smile:

PR #126, opened 2022-06-20
Upgrade ExoPlayer to 2.18

jarnedemeulemeester:
Upgrade ExoPlayer to 2.18.0.


jarnedemeulemeester:
There is a small issue with the TrackSelectionDialog.

It doesn't seem to remember what track is selected.
But other than that it works fine, it's just a visual bug.

PR #139, opened 2022-08-08
Start of chromecast support

rhysmdnz:
Added some basic chromecast support.

It's very hacky at the moment. But the basic functionality of playing something and controlling the playback is there.

It currently uses a forked version of the new unreleased jellyfin chromecast receiver app on the chromecast side. This is because the newer receiver app uses the standard google playback controls which makes it easier to implement than building everything from scratch for the custom jellyfin control protocol. Or writing a new receiver from scratch.


jarnedemeulemeester:
Nice! 

But there is a problem:
I still plan on publishing the app to F-Droid which prohibits the use of proprietary gms libraries.
And the cast library is part of gms...

The way the official Jellyfin app solves this is by creating two separate builds, one with proprietary libraries (for the Google Play Store) and one without (for F-Droid).
I think it would require a lot of work to refactor the app into two builds which is why I haven't implemented casting support yet.

hkfuertes:
Can I suggest an alternative? Maybe a companion app / plugin with just chromecast support... so that if main `findroid` app detects it, an intent can be called so that this companion app will actually cast the media? Im asuming that in the end the media is a (semi) public url...
I dont have the skils to do this, but it may solve the problem with foss...

henriquelucas:
Is there any news regarding this issue?

gerba102:
> Is there any news regarding this issue?

I like to know as well

henriquelucas:
I understand the problem with F-droid, but many who use the app experience this difficulty. Thank you so much for creating this wonderful app.

More would like to see this functionality available in Findroid

damomato:
Hopefully this can be solved as I agree it's very important. Being able to cast downloaded videos would be amazing when playback is being a bit finicky

chenweiask:
For those whom want to use this on Chromecast, maybe KODI with Jellyfin integration is the best solution so far. Or if you have a powerful server, the Jellyfin app is also a choice to use your server to decoding. Both apps could be downloaded from play store.

PR #144, opened 2022-08-21
Fix "pinch to zoom" for MPV

### LINKED ISSUE #22: add pinch to zoom
Many phones have 19:8, 21:8 ratio so it would be good to add this

Hjaelp:
Closes #22


jarnedemeulemeester:
Thanks!

PR #166, opened 2022-09-16
Fix MPV subtitles zoom in

### LINKED ISSUE #162: Subtitles cut off when zooming in with MPV
**Describe the bug**
When MPV is disabled as the player, zooming the video in to fit your screen makes the subtitles stay on the same spot of your screen but when MPV is enabled, the subtitles stay on the same spot of the video, so they're cut off.

**To Reproduce**
1. Go into Settings > Player
2. Enable "mpv player"
3. Watch something with subtitles and zoom in

**Screenshots**
- MPV disabled (subtitles where they should be):
Screenshot_20220910-204448_Findroid

- MPV enabled (subtitles cut off)
Screenshot_20220910-204536_Findroid

Hjaelp:
Closes #162
Some .ass subtitles will still be broken due to their forced stylings unless stripped.

### DIFF
--- a/app/src/main/java/dev/jdtech/jellyfin/mpv/MPVPlayer.kt
+++ b/app/src/main/java/dev/jdtech/jellyfin/mpv/MPVPlayer.kt
@@ -1210,6 +1210,15 @@ class MPVPlayer(
     fun updateZoomMode(enabled: Boolean) {
         val level = if (enabled) "1" else "0"
         MPVLib.setOptionString("panscan", level)
+
+        if (enabled) {
+            MPVLib.setOptionString("sub-use-margins", "yes")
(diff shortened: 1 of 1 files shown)


jarnedemeulemeester:
Looks good :+1: 
Will do a little bit more testing before merging.

Hjaelp:
> Looks good 👍 Will do a little bit more testing before merging.

Sure, no problem.

I think ideally there should also be an option in settings for MPV to remove all styling by setting sub-ass-override to force/strip. However these new defaults are fine for most content/subtitles.

jarnedemeulemeester:
> I think ideally there should also be an option in settings for MPV to remove all styling by setting sub-ass-override to force/strip. However these new defaults are fine for most content/subtitles.

Yeah I agree!

PR #198, opened 2022-11-18
Handle audio focus

PaulWoitaschek:
Handle the audio focus, so that findroid cooperates with other media players.
For example, findroid will now pause when the user gets a call.


jarnedemeulemeester:
Thanks a lot!

PR #230, opened 2023-01-07
Modularize the codebase

jarnedemeulemeester:
Split the codebase into multiple modules:
- `app:phone` The main entrypoint for the phone app
- `core` Core functionality used in the different `app` modules
- `player:core` Core player functionality (also contains the downloads logic) mostly used in the other `player` modules
- `player:video` Video player functionality
- `data` The Jellyfin data repository
- `preferences` App preferences

Later we will also add the following modules:
- `app:tv` The main entrypoint for the tv app
- `app:wear` The main entrypoint for the wear app
- `player:audio` Music player functionality


jarnedemeulemeester:
While I don't think this is finished yet. I think it's a good idea to merge it now so that other work can continue.
Later changes can always be made to make the modules less dependent on each other and build out a cleaner architecture. 
This effort will already allow the creation of a separate tv and wear versions without including everything from the phone version.

PR #237, opened 2023-01-15
Picture in picture

Satellite-Galaxy:
_I accidentally closed my previous PR when syncing my fork so I had to make a new one._

This PR implements picture in picture support when playing media.

I have gotten this almost fully working except for 2 bugs that I have yet to figure out how to fix.

# Current Bugs
- [x] When returning back to full screen your playback starts at where you left off before entering PIP.
- [ ] If you close PIP another instance (?) of the app is opened in the background and continues audio playback.


Satellite-Galaxy:
My bad it looks like I mistakenly reverted a commit somehow??

PR #238, opened 2023-01-19
Picture In Picture

Saturn745:
_Third times a charm..._

This PR implements picture in picture.

I have tested this and have not found any bugs myself so please let me know of any issues.

### DIFF
--- a/app/phone/src/main/AndroidManifest.xml
+++ b/app/phone/src/main/AndroidManifest.xml
@@ -18,6 +18,9 @@
 
         <activity
             android:name=".PlayerActivity"
+            android:supportsPictureInPicture="true"
+            android:configChanges=
+                "screenSize|smallestScreenSize|screenLayout|orientation"
             android:screenOrientation="sensorLandscape" />
 
         <activity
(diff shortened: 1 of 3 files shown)


Saturn745:
> Awesome work! I tested it and it works pretty good.
> 
> There is however one bug I noticed:
> When playing a video in PiP mode and returning to the app using the app switcher. The app will open up on the Home screen (while still playing the video in PiP mode). When you now start playing an other video the app crashes.
> Ideally when returning to the app using the app switcher you should return to the player and PiP mode should be stopped.
> 
> Also two more things I would like to see (but can be implemented by me in a later PR):
> - An option in the settings to disable PiP (or a redirect to the system app settings where you can disable PiP).
> - A button on the player UI that starts PiP.

I will get to fixing that and implementing those two options.

Saturn745:
> When playing a video in PiP mode and returning to the app using the app switcher. The app will open up on the Home screen (while still playing the video in PiP mode). When you now start playing an other video the app crashes.
Ideally when returning to the app using the app switcher you should return to the player and PiP mode should be stopped.


I still haven't found a way to fix this yet (not a good way at least).


> A button on the player UI that starts PiP.

I was gonna work on this but could not find a good icon to use.

Saturn745:
Seeing as #277 has been opened and has a better implementation I will be closing this PR.

PR #246, opened 2023-01-27
feat(media): add detailed metadata for file on `MediaInfoFragment`

### LINKED ISSUE #93: [Feature request] More metadata
More metadata like Video resolution, audio track and subtitles.
Excample picture is from the original jellyfin app.

Fittingly, a drop down menu for the audio track would be handy (if more than one is available).

Yash-Garg:
- Detailed information regarding A/V and Subtitles (Hidden behind a preference)
- ChipGroup for showing metadata in a minimal form (incl. Dolby logo)
- Closes #93


tam1m:
I'd like to comment on this from a UX perspective.

In my opinion there are two different perspectives as to why subtitles are listed (to an extend this might be true for audio tracks aswell).

The first perspective is the technical one. This is pretty much what every JF client and the screenshots in this implementation do. 

Here everything about the different subtitles is shown, wich for the most part is technical information. So language, type (SDH,CC,etc) and subtitle format (DVDSUB, SUBRIP, etc). 

Except for the language, the information listed is only relevant from a technical perspective, where the normal user pretty much only cares for languages available.

```
English SRT - Default - SUBRIP, English SDH SRT - SUBRIP, Danish - DVDSUB, Finnish - DVDSUB
```



The second perspective is the "normal user" one. Where one would only be interested in subtitle languages available and *maybe* subtitle type i.e. SDH, CC and so on.

This is a pretty prevalent thing in JF and many open source projects in general, where oftentimes the UI gets cluttered with too much technical information that the normal user pretty much never is interested in and also doesn't need.

So my proposable would be to implement the subtitle info as a simple list of languages and again maybe the type in brackets.

So this
```
English SRT - Default - SUBRIP, English SDH SRT - SUBRIP, Danish - DVDSUB, Finnish - DVDSUB
```
would simply become
```
ENG,ENG[SDH],DEN,FIN
```

In my opinion, this looks way cleaner, shorter, less cluttered, while still presenting all the information a typical user actually needs.
 
Keep in mind, the actual subtitle selection dialog presented to the user in the player interface, can still show all the technical information, if this is something, people want and need. Wich in my opinion is dabatable aswell, but out of scope for this issue.

Yash-Garg:
> I'd like to comment on this from a UX perspective.
> 
> In my opinion there are two different perspectives as to why subtitles are listed (to an extend this might be true for audio tracks aswell).
> 
> The first perspective is the technical one. This is pretty much what every JF client and the screenshots in this implementation do.
> 
> Here everything about the different subtitles is shown, wich for the most part is technical information. So language, type (SDH,CC,etc) and subtitle format (DVDSUB, SUBRIP, etc).
> 
> Except for the language, the information listed is only relevant from a technical perspective, where the normal user pretty much only cares for languages available.
> 
> ```
> English SRT - Default - SUBRIP, English SDH SRT - SUBRIP, Danish - DVDSUB, Finnish - DVDSUB
> ```
> 
> The second perspective is the "normal user" one. Where one would only be interested in subtitle languages available and _maybe_ subtitle type i.e. SDH, CC and so on.
> 
> This is a pretty prevalent thing in JF and many open source projects in general, where oftentimes the UI gets cluttered with too much technical information that the normal user pretty much never is interested in and also doesn't need.
> 
> So my proposable would be to implement the subtitle info as a simple list of languages and again maybe the type in brackets.
> 
> So this
> 
> ```
> English SRT - Default - SUBRIP, English SDH SRT - SUBRIP, Danish - DVDSUB, Finnish - DVDSUB
> ```
> 
> would simply become
> 
> ```
> ENG,ENG[SDH],DEN,FIN
> ```
> 
> In my opinion, this looks way cleaner, shorter, less cluttered, while still presenting all the information a typical user actually needs.
> 
> Keep in mind, the actual subtitle selection dialog presented to the user in the player interface, can still show all the technical information, if this is something, people want and need. Wich in my opinion is dabatable aswell, but out of scope for this issue.

Thanks for the suggestion @tam1m! I get your point that we need to show only minimal information for the Subtitles as well as the Audio tracks. 

Right now, I'm just using the titles which Jellyfin API gives us directly. So, it may be hard to convert them to generic shorter form for all the Subtitles that may be present in a video.

I'll try and check if it's possible to show a more simpler text for the list, which can be used for any subtitle file.

tam1m:
> Thanks for the suggestion @tam1m! I get your point that we need to show only minimal information for the Subtitles as well as the Audio tracks.

As for the audiotacks I'm not sure, as the audioformat, in my opinoin, is still pretty relevant to users. One might actually want to know if there is an 5.1 or Atmos track or whatever.

> Right now, I'm just using the titles which Jellyfin API gives us directly. So, it may be hard to convert them to generic shorter form for all the Subtitles that may be present in a video.

I don't know the Jellyfin API, so yeah, if there is no way to query for subtitle specifics like format and type, this needs to be parsed locally from the string JF returns. I don't know if that's something you want to implement for one reason or another. I just wanted to give some input on this.

Yash-Garg:
> > Thanks for the suggestion @tam1m! I get your point that we need to show only minimal information for the Subtitles as well as the Audio tracks.
> 
> As for the audiotacks I'm not sure, as the audioformat, in my opinoin, is still pretty relevant to users. One might actually want to know if there is an 5.1 or Atmos track or whatever. I still would get rid of the `DEFAULT` string in list of audiotracks though
> 
> > Right now, I'm just using the titles which Jellyfin API gives us directly. So, it may be hard to convert them to generic shorter form for all the Subtitles that may be present in a video.
> 
> I don't know the Jellyfin API, so yeah, if there is no way to query for subtitle specifics like format and type, this needs to be parsed locally from the string JF returns. I don't know if that's something you want to implement for one reason or another. I just wanted to give some input on this. 

I saw some subtitles did not contain language and some other fields which other subs have, so it may not be completely relevant.

I'll try parsing it locally on the client's end and see if it works decent for all the files. Thanks for your input!

Jcuhfehl:
I would just like to add that sometimes (mostly for anime) more subtitle information than just the language and whether it's cc or not is relevant. A dual audio anime episode probably has at least a dialogue and signs and songs track, both of which would be marked English, and there could be even more if it has multiple choices for fansubs and things like that, so I think we should at least also show the subtitle track name.

tam1m:
> I would just like to add that sometimes (mostly for anime) more subtitle information than just the language and whether it's cc or not is relevant. A dual audio anime episode probably has at least a dialogue and signs and songs track, both of which would be marked English, and there could be even more if it has multiple choices for fansubs and things like that, so I think we should at least also show the subtitle track name.

Could you provide some examples for anime subtitles? I don't watch anime and have no clue as to how they are shown in JF. This makes it a bit more complicated, I guess, but maybe those "tags" could simply be added as additional tags inside the bracket.

Jcuhfehl:
Something like
`Full (Commie) - Eng - Default - ASS`
`Signs & Songs (Commie) - Eng - ASS`
The part that is relevant here is `Full (Commie)`/`Signs & Songs (Commie)` (well, and the language of course)

Yash-Garg:
@jarnedemeulemeester what do you think? How should we handle this?

jarnedemeulemeester:
I do have to agree with @tam1m that if we show that much information it becomes a bit cluttered.

Okay so after thinking about this for some time I suggest we do the following:
- Keep the current implementation but don't show it by default. Make it an option in the settings to show it.
- Implement something like this:
image
Just show some tags in the metadata bar. That's how most streaming apps to it and I quite like it. Looks clean and provides some useful information. I understand this may not provide enough information for everyone but the more detailed metadata can be enabled in the settings.

Tags:
- Resolution: 4K, HD (gone if less than HD)
- HDR: HDR, HDR10, HLG, DOLBY VISION (gone if not HDR)
- Audio: 2.0, 5.1, 7.1, DOLBY DIGITAL, DOLBY DIGITAL+, DOLBY ATMOS, DTS, DTS-HD, DTS:X
- Subtitles: CC (shown if there are any subtitles)

(Replace DOLBY by an icon)

Also, in a later PR we could make those tags clickable so they open a popup and show more information. But I would add that later.

Yash-Garg:
> I do have to agree with @tam1m that if we show that much information it becomes a bit cluttered.
> 
> After thinking about this for some time I suggest we do the following:
> 
> * Keep the current implementation but don't show it by default. Make it an option in the settings to show it.
> * Implement something like this:
>   image
>   Just show some tags in the metadata bar. That's how most streaming apps to it and I quite like it. Looks clean and provides some useful information. I understand this may not provide enough information for everyone but the more detailed metadata can be enabled in the settings.
> 
> Tags:
> 
> * Resolution: 4K, HD (gone if less than HD)
> * HDR: HDR, HDR10, HLG, DOLBY VISION (gone if not HDR)
> * Audio: 2.0, 5.1, 7.1, DOLBY DIGITAL, DOLBY DIGITAL+, DOLBY ATMOS, DTS, DTS-HD, DTS:X
> * Subtitles: CC (shown if there are any subtitles)
> 
> (Replace DOLBY by an icon)
> 
> @Jcuhfehl I also have a lot of anime with dual audio tracks. The option to show the detailed metadata should cover that usecase.
> 
> Also, in a later PR we could make those tags clickable so they open a popup and show more information. But I would add that later.
> 
> @Yash-Garg what do you think?

I think this will look much better and clean, will just need to properly parse all the stuff though which might be flaky.

I'll try implementing it and share the progress here!

Yash-Garg:
@jarnedemeulemeester How should the chips that are exceeding the space be handled here?

image

Screenshot_20230131_161744

anunaym14:
I like it much better in the second screenshots where chips are in the next row!

Yash-Garg:
@jarnedemeulemeester I had some other questions also regarding the parsing and showing of the data related to Audio. I have a rough implementation in [`76a997d` (#246)](https://github.com/jarnedemeulemeester/findroid/pull/246/commits/76a997dd38e5c6dcf88fe39028fbafe3ba38f93b), which takes all the metadata from list of available A/V sources.

As you mentioned audio formats we show `5.1` in the mockup, but that's just the channel layout. So should we show another chip for the `Dolby Digital` audio stuff? 

Also, there can be multiple tracks so what should we prioritise to show and which track. How will that be decided for the chips?

jarnedemeulemeester:
@Yash-Garg Let's indeed move the chips to the next row.

You make a good point about channel layout vs codec. Let's always show the channel layout. And in an other chip the codec but maybe only if it's DOLBY or DTS? I don't really see the need to put `AAC`, `OPUS` or something else there. But maybe you disagree? I just don't want to make it to cluttered.
For the DOLBY logo, we could just look at the codec right? `ac3`, `eac3`, `truehd`

If there are multiple tracks I think we should just pick the default one. 

PS I haven't been able to test this yet but it seems like you're doing a great job! :)

Yash-Garg:
@jarnedemeulemeester I think `DOLBY` and `DTS` are fine, others aren't required as such.

Regarding the `DOLBY` logo, I'll check if the API gives proper codec value and implement.

I guess the default track will always be the `first()` track, so I'll use it only for showing the codec and channel layout.

Thanks!

Yash-Garg:
@jarnedemeulemeester I added the dolby logo but confused with the colors, should I match it with the M3 Dynamic Theme? Also, is the positioning fine?

neopc10:
Is it really necessary to distinguish HDR10 from HDR? HLG too, really. I'd go for HDR, HDR10+, and Dolby Vision.

Also, re: Dolby -- DD, DD+, TrueHD. There can be "Dolby Digital Plus with Dolby Atmos" and "TrueHD with Atmos" so it's worth making the distinction. I think removing "Dolby" from the chip is a good move.

Hopefully with these chips it'd minimize the need to have the Video/Audio/Subtitles info beneath all those buttons in an expanded appearance. Maybe they can be bumped all the way down and collapsed.

This is probably out of the scope of this PR but maybe round the rating score to the nearest tenth?

Yash-Garg:
> Is it really necessary to distinguish HDR10 from HDR? HLG too, really. I'd go for HDR, HDR10+, and Dolby Vision.
> 
> Also, re: Dolby -- DD, DD+, TrueHD. There can be "Dolby Digital Plus with Dolby Atmos" and "TrueHD with Atmos" so it's worth making the distinction. I think removing "Dolby" from the chip is a good move.
> 
> Hopefully with these chips it'd minimize the need to have the Video/Audio/Subtitles info beneath all those buttons in an expanded appearance. Maybe they can be bumped all the way down and collapsed.
> 
> This is probably out of the scope of this PR but maybe round the rating score to the nearest tenth?

@neopc10 We are just directly using the codec values from what Jellyfin API gives us here. HDR10/HDR10+ show the same codec I think in the API, I will recheck once with a different file.

The `DD+, with Atmos` will be hard to parse since there are no proper fields that give the information regarding it. We will need to basically parse the whole audio track title which won't be feasible in my opinion. I'll let @jarnedemeulemeester suggest what he thinks about it. 

Regarding the extra info, that will be hidden behind a setting (ignore it in the screenshots).

jarnedemeulemeester:
@neopc10 I still would like to see if it's HDR10 or HLG. I know HLG is not super popular but still. I don't think its bad to show these two by their name instead of merging them under "HDR".

@Yash-Garg 
These are the chips I would use.
image
I would also put the Dolby logo in the chip as seen.

About Atmos, we should just parse the title of the first track to see if it includes `atmos` in my opinion. This is they only way I think.
If Atmos is included in DD+ or TrueHD then don't show DD+ or TrueHD but show Atmos chip. At least that's what I would do. But on second thought we could also just add "Atmos" to the chip like "Digital Plus with Atmos", "TrueHD with Atmos" as suggested by @neopc10. I'm not sure, what do you prefer?

I'm also not sure about the DTS codecs, my knowledge about them is pretty limited. If there are any missing or wrong feel free to correct me. :)

Yash-Garg:
> @neopc10 I still would like to see if it's HDR10 or HLG. I know HLG is not super popular but still. I don't think its bad to show these two by their name instead of merging them under "HDR".
> 
> @Yash-Garg These are the chips I would use. image I would also put the Dolby logo in the chip as seen.
> 
> About Atmos, we should just parse the title of the first track to see if it includes `atmos` in my opinion. This is they only way I think. If Atmos is included in DD+ or TrueHD then don't show DD+ or TrueHD but show Atmos chip. At least that's what I would do. But on second thought we could also just add "Atmos" to the chip like "Digital Plus with Atmos", "TrueHD with Atmos" as suggested by @neopc10. I'm not sure, what do you prefer?
> 
> I'm also not sure about the DTS codecs, my knowledge about them is pretty limited. If there are any missing or wrong feel free to correct me. :)

I think showing `Digital Plus with Atmos` will be too long in a text, I will prefer showing the `Dolby Logo Atmos`.

Yash-Garg:
@jarnedemeulemeester The API gives `HDR10` and `HDR10+` as the same codec and I don't think there is a proper way to differentiate both of them. So, should I keep only `HDR10`?

jarnedemeulemeester:
@Yash-Garg Yeah I think it's okay to just keep `HDR10`.

The only think I would change about the appearance now is the height. I would reduce it a bit like in the previous screenshots.

Let me now test this... :smile:

neopc10:
@Yash-Garg Looks good, my only gripe is the distinction between TrueHD and DD+ with Atmos. Is that EAC3? Is it the uncompressed track? I think the Atmos bit needs to be an additional chip.

Yash-Garg:
> @Yash-Garg Looks good, my only gripe is the distinction between TrueHD and DD+ with Atmos. Is that EAC3? Is it the uncompressed track? I think the Atmos bit needs to be an additional chip.

I think we can do it like this.

image

Yash-Garg:
@jarnedemeulemeester You can test it out once, I've only kept `DTS` for that category since I couldn't find any codec in the API for it.

neopc10:
> I think we can show it like this.

Yes much better!

jarnedemeulemeester:
I changed it so that it only parses metadata for movies. No reason to do that work for series.
And also corrected the spacing when there a no chips.

I will test with a few more media files before merging.

Yash-Garg:
> I changed it so that it only parses metadata for movies. No reason to do that work for series. And also corrected the spacing when there a no chips.
> 
> I will test with a few more media files before merging.

Great!

Yash-Garg:
I noticed an issue with the DoVi parsing @jarnedemeulemeester, I'll push a fix for it after checking some stuff!

Yash-Garg:
I did not find any other issues as such, so if you have completed testing your side @jarnedemeulemeester it's ready for merging.

neopc10:
> I changed it so that it only parses metadata for movies. No reason to do that work for series. And also corrected the spacing when there a no chips.

Wouldn't you want to parse for individual episodes?

Yash-Garg:
> > I changed it so that it only parses metadata for movies. No reason to do that work for series. And also corrected the spacing when there a no chips.
> 
> Wouldn't you want to parse for individual episodes?

The page for individual episodes isn't the same as of now and we currently only show it for movies, so kind pointless to unnecessarily parse it. It can be done later when if he wants to add a whole page for the TV episodes.

data/src/main/java/dev/jdtech/jellyfin/models/VideoMetadata.kt
```diff
@@ -0,0 +1,46 @@
...
+    CH_5_1("5.1"),
+    CH_7_1("7.1");
+}
+
+enum class AudioCodec(val raw: String) {
+    FLAC("FLAC"),
+    MP3("MP3"),
+    AAC("AAC"),
+    AC3("Digital"),
+    EAC3("DD+"),
```
jarnedemeulemeester, line 39:
```suggestion
    EAC3("Digital+"),

```

Or `Digital Plus` but maybe that's too long.

data/src/main/java/dev/jdtech/jellyfin/models/VideoMetadata.kt
```diff
@@ -0,0 +1,46 @@
...
+    VORBIS("VORBIS"),
+    DTS("DTS"),
+    TRUEHD("TrueHD"),
+    OPUS("OPUS");
+
+    override fun toString() = this.raw.lowercase()
```
jarnedemeulemeester, line 45: Dolby Digital and Digital Plus are never shown. I think it is because you are comparing the codec from Jellyfin with `raw` but you should compare it to the lowercase enum. But maybe then the name `raw` does not make sense anymore? :thinking: 

This works:

```suggestion
    override fun toString() = super.toString().lowercase()

```

jarnedemeulemeester (changes requested):
I think this is almost ready to be merged :smile:

data/src/main/java/dev/jdtech/jellyfin/models/VideoMetadata.kt
Yash-Garg: `Digital+` will be better I think!

data/src/main/java/dev/jdtech/jellyfin/models/VideoMetadata.kt
Yash-Garg: Yup, I was skeptical about codecs and there is no list specified in the Jellyfin API docs. I think for `Dolby Digital` and `Dolby Digital+` we will have to parse it like we did for `Atmos`? What do you think @jarnedemeulemeester?

Yash-Garg:
Works fine now! @jarnedemeulemeester 

image

jarnedemeulemeester (approved):

PR #247, opened 2023-01-27
Fix mpv in downloads

Jcuhfehl:
With this pr, users don't have to click the play button when playing from downloads in mpv. This is achieved by adding a 1ms delay just before the `player.prepare()` function is called. For some reason, this ensures it doesn't crash when that function is called, so we can now call it, fixing having to manually press play.


jarnedemeulemeester:
As much as I don't like this "fix". It does work so I'll merge it :sweat_smile: 
Ultimately I really need to work on the andriodx media3 player -> libmpv implementation, I suspect a lot of bugs like this can be fixed there.

Jcuhfehl:
Yeah, I really agree with you on it being very hacky. Let's hope that media3 actually fixes this in a proper way!

PR #257, opened 2023-01-31
Fix resuming streamed content

### LINKED ISSUE #256: Resuming playback doesn't work with with streamed content
**Describe the bug**
Resuming playback does not work for streamed content. The video just starts from the beginning.

**To Reproduce**
1. Play any video with playback progress.

**Expected behavior**
Continues playing from the last playback position.

 - Device: Google Pixel 6
 - Android version: 13
 - App version: v0.10.0 (16)
 - Jellyfin version: [e.g. 10.8.9]

**Additional context**
The cause seems to be commit https://github.com/jarnedemeulemeester/findroid/commit/12e743afcd86b8701386c11d58d10bb941202991# where the playback progress is divided by 10000 for all content (should probably be only downloads)

Jcuhfehl:
Fixes #256 by dividing playback progress by 10000 when the `PlayerItem` for downloads is created, like it is for streamed content, so that the division can be removed when starting playback.


jarnedemeulemeester:
Thanks for the quick fix!

PR #258, opened 2023-01-31
Add season downloads

### LINKED ISSUE #73: Downloader suggestion
Allow us to download full series or seasons

Jcuhfehl:
This pr adds a button to the season view that downloads the entire season at once. Closes #73 
Things that still need to be done:
- [x] Check if it's possible to replace the usage of `GlobalScope` while still downloading the entire season if the user leaves the season fragment
- [x] (?) Add better UI feedback when clicking downloads (hoping for suggestions)
- [x] Add a button to delete all downloads of a series (because I got annoyed manually deleting downloads when testing)


Jcuhfehl:
I decided that the UI feedback is good enough already, so it's ready for review now.

Jcuhfehl:
I decided that the UI feedback is good enough already, so it's ready for review now.

player/core/src/main/java/dev/jdtech/jellyfin/utils/DownloadUtilities.kt
```diff
@@ -110,12 +109,13 @@ fun isItemAvailable(itemId: UUID): Boolean {
 fun canRetryDownload(itemId: UUID, downloadDatabaseDao: DownloadDatabaseDao, context: Context): Boolean {
     if (isItemAvailable(itemId))
         return false
-    val downloadId = downloadDatabaseDao.loadItem(itemId)?.downloadId ?: return false
+    val downloadId = downloadDatabaseDao.loadItem(itemId)?.downloadId ?: return true
```
AChep, line 112: Is this intended? Does not look quite right to retry if the item is to in the database.

Jcuhfehl:
Well, this pr doesn't really work with the new system, so I'll close it.

PR #268, opened 2023-02-03
feat(ui): make preference dialogs and switches material3-esque

Yash-Garg:
Since the app is built around Material3, we can use overlays as of now for `androidx.preferences`. Also, the `ListPreference` looks much cleaner and modern for selections with Material3.


jarnedemeulemeester:
Nice! I didn't know that was possible. 
It indeed looks way more modern now. I hated that the settings were using the old material look.

Tested and works perfectly.

While I do plan on switching to compose someday. This is a nice improvement in the meantime. :smile:

Yash-Garg:
> Nice! I didn't know that was possible. It indeed looks way more modern now. I hated that the settings were using the old material look.
> 
> Tested and works perfectly.
> 
> While I do plan on switching to compose someday. This is a nice improvement in the meantime. 😄

Thanks! Also, sadly no library for compose is there for preferences :/ will have to build either settings completely manually or just rely on the `androidx.preferences`. I hope they release something in future!

PR #270, opened 2023-02-04
refactor(ui): use a global preference style for Material3 switch

Yash-Garg:
Optimising #268 even more, we can define a custom style for the preference so that we don't have to add `widgetLayout` everytime!


jarnedemeulemeester:
Nice!

Issue #210, opened 2022-12-04
Downloads failed on multiple devices


AnonymousWebHacker:
Device: Redmi 9A , Android 11 2GB RAM
On my device it works perfect, I can download media without problems. I press the download button and the system download manager takes care of the rest.

The same does not happen with other household cell phones and tablets. When pressing the download button, it simply does nothing, a download manager does not appear, no notification appears, or anything.
The error happens in other Redmi android 10, Redmi android 11 the same, samsung A7, android 7 ..etc

It is a common error in many devices https://github.com/jarnedemeulemeester/findroid/issues/87#issuecomment-1325626164

Jcuhfehl:
I just mentioned it in the other issue, but you could try to manually enable the files and media permission. Also, if you know how, getting logs from the app might also help.

AnonymousWebHacker:
It could be app permissions issues, let me check and I'll let you know

AnonymousWebHacker:
> Lo acabo de mencionar en el otro problema, pero puede intentar habilitar manualmente los archivos y el permiso de medios. Además, si sabe cómo, obtener registros de la aplicación también podría ayudar.

After reviewing several times, I realize that if I downloaded any media from the browser, it downloads fine, but when trying to use the download manager, it doesn't allow downloading (this is over Wi-Fi and without internet access)

The problem is not with findroid, it is with the download manager, which in device settings, APP DOWNLOADS, has access to data, but not Wi-Fi

MudssirAhmed:
I start working on this issue and let you know the updates.
@jarnedemeulemeester If possible you can assign it to me.

MudssirAhmed:
@AnonymousWebHacker Can you share screen recording or steps to reproduce this issue.

MudssirAhmed:
The first time I downloaded it was working at my side and the download icon will changed to red. And I cancel the download and hit the download button again then the download thing is not happening. 


@AnonymousWebHacker is this is your issue.

@jarnedemeulemeester  is this the right flow or we need to change it?

AnonymousWebHacker:
> The first time I downloaded it was working at my side and the download icon will changed to red. And I cancel the download and hit the download button again then the download thing is not happening.
> 
> @AnonymousWebHacker is this is your issue.
> 
> @jarnedemeulemeester is this the right flow or we need to change it?

The download option has 2 errors.

[...]

PR #274, opened 2023-02-05
Download handled is it exist in DB but not in file-storage

MudssirAhmed:
If the user clicks on the download button and if the media do not exist in File-System but exist it DB. Then it will delete the previous DB entry and create a new one and start the download.

[#210 ](https://github.com/jarnedemeulemeester/findroid/issues/210)

### DIFF
--- a/player/core/src/main/java/dev/jdtech/jellyfin/utils/DownloadUtilities.kt
+++ b/player/core/src/main/java/dev/jdtech/jellyfin/utils/DownloadUtilities.kt
@@ -122,7 +122,10 @@ fun canRetryDownload(itemId: UUID, downloadDatabaseDao: DownloadDatabaseDao, con
 
 fun isItemDownloaded(downloadDatabaseDao: DownloadDatabaseDao, itemId: UUID): Boolean {
     val item = downloadDatabaseDao.loadItem(itemId)
-    return item != null
+    if (item != null) {
+        return File(defaultStorage, item.id.toString()).exists() && File(defaultStorage, "${item.id}.downloading").exists()
+    }
+    return false
 }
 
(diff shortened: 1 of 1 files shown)


Jcuhfehl:
While I think we need a way to indicate in the `EpisodeBottomViewModel`,

Jcuhfehl:
While I do think we need a way to show download status in the episode bottom sheet and mediainfofragment, I don't think this is the way, because the checks you added to `isItemDownloaded` are already implemented in `canRetryDownload`. So I think it would be better to have the download button show whether the download is still ongoing, is finished, or needs to be retried by using the functions that already exist. This way, it also doesn't look like the download just randomly dissappeared, but makes it clear that the download has failed and needs to be attempted again.

MudssirAhmed:
Ya! I think you are pointing out the right points, can you suggest the way, how I implement the UI.

Can I add icons that show, the download status, or something else.

Jcuhfehl:
There is some code that does basically the thing we want in the `MediaInfoFragment` and `EpisodeBottomSheetFragment` in the section that deals with the downloaded files. The harder question is how we make the UI for this look, because if we replace the download button with a retry button when the download failed it's kinda ambiguous and could be confusing. Honestly, I don't have the answer to what's best in this case.

MudssirAhmed:
Ya that's why I think the current flow is best for download, but when use hits the download button again and the file is already downloaded, we can show a tost or a pop-up that indicates the file is already downloaded.

MudssirAhmed:
Hey @Jcuhfehl, I think we can add Icons with text. There are two versions:-

1.  (Download-Icon) with (Download-Text)
2. (Completed-Icon) with (Downloaded-Text)

And I think we do have not any requirement to show the Download-Error.

PR #282, opened 2023-02-09
Add AMOLED dark theme

zx900930:
Colors changed to pure black:
`android:colorBackground`
`colorSurface`

For comparison:
Screenshot_2023-02-09-17-21-55-1063182239
Screenshot_2023-02-09-17-22-06-1763558161

Knowing issues:
- [x] ~~The dynamic colors option wont work if you activated this option~~.


app/phone/src/main/java/dev/jdtech/jellyfin/fragments/SettingsAppearanceFragment.kt
```diff
@@ -19,5 +21,13 @@ class SettingsAppearanceFragment : PreferenceFragmentCompat() {
             }
             true
         }
+
+        findPreference<SwitchPreferenceCompat>("pref_amoled_theme")?.setOnPreferenceChangeListener { _, isAmoledTheme ->
+            when (isAmoledTheme) {
+               true -> requireActivity().recreate()
+               false -> requireActivity().recreate()
+            }
```
Yash-Garg, line 29: We can just directly use this, instead of using a switch-case.

```suggestion
            requireActivity().recreate()
```

Yash-Garg (changes requested):

app/phone/src/main/java/dev/jdtech/jellyfin/fragments/SettingsAppearanceFragment.kt
zx900930: Done, I was using a diffrent approach before and forgot to change this part ☺

app/phone/src/main/java/dev/jdtech/jellyfin/MainActivity.kt
```diff
@@ -38,6 +38,12 @@ class MainActivity : AppCompatActivity() {
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
 
+        if (appPreferences.amoledTheme) {
+            setTheme(R.style.Theme_FindroidAMOLED)
+        } else {
+            setTheme(R.style.Theme_Findroid)
+        }
```
jarnedemeulemeester, line 45: Else statement seems to be unnecessary so I think this can be removed.

preferences/src/main/java/dev/jdtech/jellyfin/AppPreferences.kt
```diff
@@ -22,8 +22,9 @@ constructor(
         }
 
     // Appearance
-    val theme get() = sharedPreferences.getString(Constants.PREF_THEME, null)
+    val theme: String? get() = sharedPreferences.getString(Constants.PREF_THEME, null)
```
jarnedemeulemeester, line 25: I would not add `: String?` as the type is already inferred from the getter.
This was probably added during testing?

jarnedemeulemeester (changes requested):

preferences/src/main/java/dev/jdtech/jellyfin/AppPreferences.kt
zx900930: yeap, I think I already removed it, and after running the test, it appears automatically again.

app/phone/src/main/java/dev/jdtech/jellyfin/MainActivity.kt
zx900930: done

jarnedemeulemeester (approved):

PR #286, opened 2023-02-12
fix: Implementation of Mandatory System Gesture Insets to Avoid Conflicts with System Gestures

### LINKED ISSUE #281: Feature: add zones for gestures
Currently with the mpv player, it is almost imporsible to use android's gesture based navigation because the player recognize the gesture itself. Thus when I try to go back to the movie info I get the player to change the payback position. It would be great to recognize gestures only in the middle of the screen and leave the borders to the system so navigation gestures work properly.

anilbeesetti:
- Previously, the use of system gestures (such as the back and home gestures) resulted in the invocation of app gestures (such as seek, volume and brightness gestures), causing conflicts with the system gestures.
- This PR solves the issue of conflicting gestures by adding mandatory system gesture insets. These insets represent the areas of the window where system gestures have priority. This implementation is in line with the guidelines provided by the Android Developers website. And also this closes #281

### DIFF
--- a/app/phone/src/main/java/dev/jdtech/jellyfin/utils/PlayerGestureHelper.kt
+++ b/app/phone/src/main/java/dev/jdtech/jellyfin/utils/PlayerGestureHelper.kt
@@ -1,13 +1,16 @@
 package dev.jdtech.jellyfin.utils
 
 import android.annotation.SuppressLint
+import android.content.res.Resources
 import android.media.AudioManager
+import android.os.Build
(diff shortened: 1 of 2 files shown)


jarnedemeulemeester:
@anilbeesetti Nice work!

I made a few small changes. Could you please test it again and if everything is fine I will merge it.

anilbeesetti:
Great, I tested and everything is functioning as expected

PR #300, opened 2023-02-20
feat: add option to choose per-app language

Yash-Garg:
Implements functionality for https://developer.android.com/guide/topics/resources/app-languages for Android 13.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -36,5 +36,5 @@ tasks.create<Delete>("clean") {
     delete(rootProject.buildDir)
 }
 
-val appVersionCode by extra { 17 }
-val appVersionName by extra { "0.10.1" }
+val appVersionCode by extra { 18 }
+val appVersionName by extra { "0.11.0" }


Yash-Garg:
@jarnedemeulemeester Should I keep the in-app language picker? It doesn't sync the language if you change it through system in A13 but works good on all Android versions if changed from within the app.

jarnedemeulemeester:
@Yash-Garg sorry for the delay

Thinking about this I prefer the following:
- Make the setting redirect to the Android 13+ app language settings. There is an intent for that but I can't find it right now.
- Don't show the option on Android <13

Not showing the option on versions before Android 13 may not be possible with the current preferences library tho...
But I plan on completely reworking the settings in the future to better integrate with Jellyfin's user settings stored on the server.

Only showing the option to Android 13+ users means that about half of the userbase (stats form Play Console) will see the option which I find acceptable.
Plus no extra code would need to be maintained for previous Android versions.

Yash-Garg:
> @Yash-Garg sorry for the delay
> 
> Thinking about this I prefer the following:
> 
> * Make the setting redirect to the Android 13+ app language settings. There is an intent for that but I can't find it right now.
> * Don't show the option on Android <13
> 
> Not showing the option on versions before Android 13 may not be possible with the current preferences library tho... But I plan on completely reworking the settings in the future to better integrate with Jellyfin's user settings stored on the server.
> 
> Only showing the option to Android 13+ users means that about half of the userbase (stats form Play Console) will see the option which I find acceptable. Plus no extra code would need to be maintained for previous Android versions.

Thanks for the insight, I will take a look at this in the upcoming days!

PR #302, opened 2023-02-21
refactor: use `WindowInsetsControllerCompat` instead of deprecated methods

anilbeesetti:
- Use WindowInsetsControllerCompat instead of deprecated methods.
- Add a fix for WindowInsetsControllerCompat's bug which makes the action bar reappear

### DIFF
--- a/app/phone/src/main/AndroidManifest.xml
+++ b/app/phone/src/main/AndroidManifest.xml
@@ -25,7 +25,8 @@
         <activity
             android:name=".PlayerActivity"
             android:configChanges="screenSize|smallestScreenSize|screenLayout|orientation"
-            android:screenOrientation="sensorLandscape" />
+            android:screenOrientation="sensorLandscape"
+            android:theme="@style/Theme.Findroid.Player" />
 
         <activity
             android:name=".MainActivity"
--- a/app/phone/src/main/java/dev/jdtech/jellyfin/BasePlayerActivity.kt
+++ b/app/phone/src/main/java/dev/jdtech/jellyfin/BasePlayerActivity.kt
@@ -1,5 +1,6 @@
 package dev.jdtech.jellyfin
 
+import android.os.Bundle
 import android.view.View
 import android.view.WindowManager
 import androidx.appcompat.app.AppCompatActivity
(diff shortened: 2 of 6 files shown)


jarnedemeulemeester:
Hi @anilbeesetti 
Sorry for the extremely long wait...

I tested the changes but the system bars still appear when opening a dialog (like the speed selection dialog). They also don't go away anymore.

I saw you made Next Player, it looks really good! I tested it and it does not seem to have this issue.
Would you like to help me find a solution for the system bars in Findroid?

anilbeesetti:
@jarnedemeulemeester These changes working fine on my phone. System bars don't appear for me when opening dialog. I don't know why that's not the case on your phone.

jarnedemeulemeester:
@anilbeesetti Sorry I forgot to mention that this only happens on older Android devices. I tested with an Android 9 emulator and I had the problem. 
Meanwhile your app does not have this issue on the same emulator :thinking:

anilbeesetti:
The last commit fixes bars stay visible after closing dialog

anilbeesetti:
But the bars are white, which cause distraction. By adding this `WindowCompat.setDecorFitsSystemWindows(window, false)` liine we can make bars trasparent. what do you think?

jarnedemeulemeester:
@anilbeesetti OMG I finally figures it out!
When using `WindowCompat.setDecorFitsSystemWindows(window, false)` like you suggested to make the the window draw under the systembars, and setting both the navigationBar and statusBar to Transparent everything works as it should.
The key takeaway is that you have to make the bars transparent in order for them to disappear again. (If only this was documented somewhere...) It even removes the need for your last commit :smile: 

Just adding this to BasePlayerActivity solves it.
```kotlin
override fun onCreate(savedInstanceState: Bundle?) {
    super.onCreate(savedInstanceState)

    WindowCompat.setDecorFitsSystemWindows(window, false)
    window.navigationBarColor = Color.TRANSPARENT
    window.statusBarColor = Color.TRANSPARENT
}
```
Setting the navigationBar and statusBar colors could also probably be done in `themes.xml`.

jarnedemeulemeester:
Could you explain what the bug is that makes the actionbars reappear after changing the brightness? 
None of the devices and emulators I've tested (with your fix removed) seem to have this issue.

anilbeesetti:
> Could you explain what the bug is that makes the actionbars reappear after changing the brightness? None of the devices and emulators I've tested (with your fix removed) seem to have this issue.

From what i've tested, this  problem only exists on my phone, which is Redmi note 4 running lineage os android 11

checkout this video for reference:- 

https://github.com/jarnedemeulemeester/findroid/assets/66936649/2debbd3b-b535-4c83-bce0-acfb11b52593

jarnedemeulemeester:
While I love LineageOS for it's ability to extend a devices lifetime beyond what the manufacturer offers (I have also used it), I don't think I want to include fixes if they only apply to that specific OS.
If other Android versions form major manufacturers would also need this fix than it may be added again later.

Issue #361, opened 2023-05-13
Dynamic color changes in OLED theme 


Mylloon:
**Describe the bug**
Dynamic color changes when OLED theme is enabled.

**To Reproduce**
1. Have Dynamic colors enabled
2. Change to OLED theme
3. Your color changed

**Expected behavior**
The accent color shouldn't change.

**Screenshots**
| Without OLED  | With OLED |
|-------------------------|-------------------|
| ![Screenshot_2023-05-13-08-48-50-333_dev jdtech jellyfin](https://github.com/jarnedemeulemeester/findroid/assets/29067904/d0a7ed67-f9a9-4761-a263-18faaf425459) | ![Screenshot_2023-05-13-08-48-53-577_dev jdtech jellyfin](https://github.com/jarnedemeulemeester/findroid/assets/29067904/a4faf486-081e-4eb2-9085-c750cac8ffba) |

**Device info:**
 - Device: Mi Note 10 Lite
 - Android version: 12
 - App version: 0.11.0
 - Jellyfin version: 10.8.5

PR #362, opened 2023-05-13
fix: amoled theme not taking M3 colors

Yash-Garg:
Closes https://github.com/jarnedemeulemeester/findroid/issues/361


jarnedemeulemeester:
Thanks!

PR #372, opened 2023-05-17
feat: lock player controls

### LINKED ISSUE #70: Screen lock
First of all, great job with the app!

It would be nice to be able to lock the screen the exact same way as the Netflix Android app.

Hopefully not something too troublesome to implement. Sometimes my kid watches videos on my phone but not being able to lock the screen can cause problems.

AstroTheAstronaut:
**Add Lock Player feature.**

This feature is similar to the feature found in the stock Jellyfin Android app and in the VLC app.

When enabled, it disables the gestures (seek, volume and brightness control), and the default player controls, leaving only two buttons: a back button and an "unlock" button. Pressing the unlock button reverts the player back to the initial state, e.g. gestures enabled and default buttons shown.

Works with **ExoPlayer** and **MPV Player**.

Let me know of any issues.

Screenshots:
**Normal view:**

normal_controls

**Locked view**

locked_controls

Closes #70


jarnedemeulemeester:
Thanks for implementing the player locking feature!

I did find one bug:
- When locking the player and using the back button. The next video you play will still think that the player is locked (no gestures will work) but the normal UI is shown. When you then press the lock button the gestures become available again and the UI stays the same.  Further presses on the lock button work as expected.

AstroTheAstronaut:
> Thanks for implementing the player locking feature!
> 
> I did find one bug:
> 
> * When locking the player and using the back button. The next video you play will still think that the player is locked (no gestures will work) but the normal UI is shown. When you then press the lock button the gestures become available again and the UI stays the same.  Further presses on the lock button work as expected.


Yes, I noticed that as well however I though it was just some sort of temporary thing with the app, since, after restarting, it worked fine. 

I fixed that (needed to add just one line) and also removed some of the redundant line spaces and some other formatting.

Let me know if you find any other bug when it comes to this!

app/phone/src/main/java/dev/jdtech/jellyfin/PlayerActivity.kt
```diff
@@ -120,6 +134,75 @@ class PlayerActivity : BasePlayerActivity() {
...
+                        isControlsLocked = false
+                    } else {
+                        exoPlayerControlView.visibility = View.GONE
+                        lockedLayout.visibility = View.VISIBLE
+                        requestedOrientation = ActivityInfo.SCREEN_ORIENTATION_LOCKED
+                        isControlsLocked = true
+                    }
+                }
+            }
+        }
```
jarnedemeulemeester, line 204: Couldn't this whole block be replace by something much shorter:

```suggestion
        lockButton.setOnClickListener {
            exoPlayerControlView.visibility = View.GONE
            lockedLayout.visibility = View.VISIBLE
            requestedOrientation = ActivityInfo.SCREEN_ORIENTATION_LOCKED
            isControlsLocked = true
        }

        unlockButton.setOnClickListener {
            exoPlayerControlView.visibility = View.VISIBLE
            lockedLayout.visibility = View.GONE
            requestedOrientation = ActivityInfo.SCREEN_ORIENTATION_SENSOR_LANDSCAPE
            isControlsLocked = false    
        }
```
It does not seem necessary to check what type of player is used and if the controls are locked. The lock / unlock button is only shown when unlocked / locked.

jarnedemeulemeester (changes requested):

app/phone/src/main/java/dev/jdtech/jellyfin/PlayerActivity.kt
AstroTheAstronaut: I've been meaning to test such a change, but never got the time to properly do it.
About the type of player: I wasn't sure how you handle such things, so I added separate entries for both just to be sure.
I'll change this asap.

app/phone/src/main/java/dev/jdtech/jellyfin/PlayerActivity.kt
```diff
@@ -29,6 +31,8 @@ import dev.jdtech.jellyfin.viewmodels.PlayerActivityViewModel
 import javax.inject.Inject
 import timber.log.Timber
 
+var isControlsLocked: Boolean = false
```
Yash-Garg, line 34: Why not move this inside the activity only?

```suggestion
```

jarnedemeulemeester (approved):

PR #377, opened 2023-05-24
feat: add volume and brightness change animation

thofx:
when volume or brightness change the icon will be change


jarnedemeulemeester:
Thanks for this nice little addition!

PR #380, opened 2023-05-25
feat: ability to change number of covers in libraries

AstroTheAstronaut:
**Add ability to change number of movies/shows covers in the libraries sections.**

By default, for the libraries, Findroid uses a grid that is made of only 2 rows. This works fine if you don't have a lot of movies in your library, however it becomes a burden to navigate once you have a big library with a lot of movies and shows.
This feature adds a SeekBar in the "Appearance" section in the preference menu, which allows the user to change the number of rows (minimum 1, maximum 5). This changes the number of rows in the RecyclerView in both the Shows and Movies section.

**Screenshots:**

**_Default (2 rows) left, new (3 rows) right_**
default 2 new 3

Settings Menu:

settings

**Possible things to add:**

- Vibrations when changing the value of SeekBarPreference
- 3 rows instead of 2 by default
- Change the size of the cover art when changing the value of the SeekBar (when selecting 2, make the cover art smaller, when selecting 3, make them bigger, etc...)

Let me know of any bugs.

### DIFF
--- a/preferences/src/main/java/dev/jdtech/jellyfin/AppPreferences.kt
+++ b/preferences/src/main/java/dev/jdtech/jellyfin/AppPreferences.kt
@@ -34,6 +34,7 @@ constructor(
     val theme get() = sharedPreferences.getString(Constants.PREF_THEME, null)
     val dynamicColors get() = sharedPreferences.getBoolean(Constants.PREF_DYNAMIC_COLORS, true)
     val amoledTheme get() = sharedPreferences.getBoolean(Constants.PREF_AMOLED_THEME, false)
+    val spanCount get() = sharedPreferences.getInt(Constants.PREF_SPAN_NO, 2)
     var displayExtraInfo: Boolean
         get() = sharedPreferences.getBoolean(Constants.PREF_DISPLAY_EXTRA_INFO, false)
         set(value) {
--- a/app/phone/src/main/java/dev/jdtech/jellyfin/fragments/LibraryFragment.kt
+++ b/app/phone/src/main/java/dev/jdtech/jellyfin/fragments/LibraryFragment.kt
@@ -18,6 +18,7 @@ import androidx.lifecycle.repeatOnLifecycle
 import androidx.navigation.fragment.findNavController
 import androidx.navigation.fragment.navArgs
 import androidx.paging.LoadState
+import androidx.recyclerview.widget.GridLayoutManager
 import dagger.hilt.android.AndroidEntryPoint
 import dev.jdtech.jellyfin.AppPreferences
 import dev.jdtech.jellyfin.adapters.ViewItemPagingAdapter
@@ -61,6 +62,8 @@ class LibraryFragment : Fragment() {
 
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
         super.onViewCreated(view, savedInstanceState)
+
+        binding.itemsRecyclerView.layoutManager = GridLayoutManager(context, preferences.spanCount)
 
         val menuHost: MenuHost = requireActivity()
         menuHost.addMenuProvider(
--- a/core/src/main/res/xml/fragment_settings_appearance.xml
+++ b/core/src/main/res/xml/fragment_settings_appearance.xml
@@ -1,5 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
-<PreferenceScreen xmlns:app="http://schemas.android.com/apk/res-auto">
+<PreferenceScreen xmlns:app="http://schemas.android.com/apk/res-auto"
+    xmlns:android="http://schemas.android.com/apk/res/android">
     <ListPreference
         app:defaultValue="system"
         app:entries="@array/themes"
@@ -22,4 +23,12 @@
         app:key="pref_display_extra_info"
         app:summary="@string/extra_info_summary"
         app:title="@string/extra_info" />
+    <SeekBarPreference
+        app:defaultValue="2"
+        app:key="pref_span_count"
+        android:max="5"
+        app:seekBarIncrement="1"
+        app:min="1"
+        app:showSeekBarValue="true"
+        app:title="Number of rows in libraries view" />
 </PreferenceScreen>
--- a/preferences/src/main/java/dev/jdtech/jellyfin/Constants.kt
+++ b/preferences/src/main/java/dev/jdtech/jellyfin/Constants.kt
@@ -34,6 +34,7 @@ object Constants {
     const val PREF_THEME = "theme"
     const val PREF_DYNAMIC_COLORS = "dynamic_colors"
     const val PREF_AMOLED_THEME = "pref_amoled_theme"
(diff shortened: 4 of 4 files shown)


preferences/src/main/java/dev/jdtech/jellyfin/AppPreferences.kt
```diff
@@ -34,6 +34,7 @@ constructor(
     val theme get() = sharedPreferences.getString(Constants.PREF_THEME, null)
     val dynamicColors get() = sharedPreferences.getBoolean(Constants.PREF_DYNAMIC_COLORS, true)
     val amoledTheme get() = sharedPreferences.getBoolean(Constants.PREF_AMOLED_THEME, false)
+    val spanCount get() = sharedPreferences.getInt(Constants.PREF_SPAN_NO, 2)
```
Yash-Garg, line 37: Don't hardcode the value here, the dimens.xml contains a `library_columns` integer value which is used for different screen sizes as default.

https://github.com/jarnedemeulemeester/findroid/blob/5aad3781320ab7f989c54cd48a8638b3b88aa72c/core/src/main/res/values/dimens.xml#L9

https://github.com/jarnedemeulemeester/findroid/blob/5aad3781320ab7f989c54cd48a8638b3b88aa72c/core/src/main/res/values-sw720dp/dimens.xml#L4

https://github.com/jarnedemeulemeester/findroid/blob/5aad3781320ab7f989c54cd48a8638b3b88aa72c/core/src/main/res/values-sw600dp/dimens.xml#L6

Yash-Garg, line 37: Don't hardcode the value here, the `dimens.xml` file contains a `library_columns` integer value which is used for different screen sizes as default.

https://github.com/jarnedemeulemeester/findroid/blob/5aad3781320ab7f989c54cd48a8638b3b88aa72c/core/src/main/res/values/dimens.xml#L9

https://github.com/jarnedemeulemeester/findroid/blob/5aad3781320ab7f989c54cd48a8638b3b88aa72c/core/src/main/res/values-sw600dp/dimens.xml#L6

https://github.com/jarnedemeulemeester/findroid/blob/5aad3781320ab7f989c54cd48a8638b3b88aa72c/core/src/main/res/values-sw720dp/dimens.xml#L4

Yash-Garg (changes requested):

preferences/src/main/java/dev/jdtech/jellyfin/AppPreferences.kt
Yash-Garg, line 37: Don't hardcode the value here, the `dimens.xml` file contains a `library_columns` integer value which is used for different screen sizes as default.

https://github.com/jarnedemeulemeester/findroid/blob/5aad3781320ab7f989c54cd48a8638b3b88aa72c/core/src/main/res/values/dimens.xml#L9

Also, how about adding this number preference to the kebab menu? I think it will be better there!

AstroTheAstronaut: Thank you! 
I didn't know of the integer value in `dimens.xml`, my bad!
Also, by kebab menu, you mean the 3 dots in the upper right corner, in each library? 
If that's what you mean, I didn't add it there because, since there are multiple libraries, I think it makes more sense for it to be a global settings (Although I don't think it would necessarily be a bad idea to make it a per-library setting).

Yash-Garg: Yes, I meant the three dot menu. 

Also, we can keep only one single preference instead of per library since there may be too many of them, it'll just be more convenient to switch it without leaving the library in my opinion.

AstroTheAstronaut: I agree with you, but I'm assuming that app only has one library, so changing it there makes more sense.

Since Findroid can have multiple libraries, I don't think having a per-library setting to do this makes much sense, since it would only complicate things for no real reason. 

Also, especially on devices with larger screens (Tablets and especially TVs), having only 2 covers shown, regardless of the number of items in a given library, isn't really ideal, since you'd end up with massive cover images that take up the whole screen.

Thanks for the input, though!

Yash-Garg: Yup, maybe we can add landscape and portrait mode seekbar like in the screenshot above, in settings only. It will be better I think for both orientations.

AstroTheAstronaut: > Yup, maybe we can add landscape and portrait mode seekbar like in the screenshot above, in settings and use it accordingly with respect to the orientation. It will be better I think.

I'm sorry but I'm not sure I understand what you mean.
We should add a seekbar that shows up in the kebab menu when the app is in landscape mode?

Yash-Garg: > > Yup, maybe we can add landscape and portrait mode seekbar like in the screenshot above, in settings and use it accordingly with respect to the orientation. It will be better I think.
> 
> I'm sorry but I'm not sure I understand what you mean. We should add a seekbar that shows up in the kebab menu when the app is in landscape mode?

No, I meant a seekbar which changes both portrait and landscape values at once like in the screenshot `Portrait: 4 - Landscape: 6`

AstroTheAstronaut: > > > Yup, maybe we can add landscape and portrait mode seekbar like in the screenshot above, in settings and use it accordingly with respect to the orientation. It will be better I think.
> > 
> > I'm sorry but I'm not sure I understand what you mean. We should add a seekbar that shows up in the kebab menu when the app is in landscape mode?
> 
> No, I meant a seekbar which changes both portrait and landscape values at once like in the screenshot `Portrait: 4 - Landscape: 6`, since we would want more rows while in landscape mode.

Oh, so individual preferences for each screen orientation.

I like the idea, but I don't think seekbars would be appropriate for this, as it would look quite odd.

Also, another concern I have is that it would complicate things for the end user quite a lot more.

AstroTheAstronaut:
I'm ngl, I'm not sure how to fix the gradle stuff. Any help would be appreciated.

Yash-Garg:
> I'm ngl, I'm not sure how to fix the gradle stuff. Any help would be appreciated.

Will need to move the strings to a separate module, `implementation(project(:core))` won't work as it will cause cyclic dependency since `:core`and `:preferences` will depend on each other.

What do you think @jarnedemeulemeester ?

gerba102:
Any update on feature?

AstroTheAstronaut:
> Any update?

There is work that needs to be done to implement this properly, besides needing to update the code for other reasons, and I don't currently the time to do this because of exams.

jarnedemeulemeester:
No longer relevant since the compose rework

PR #401, opened 2023-06-06
feat: add double tap ripple animation

mustafadakhel:
The double tap seeking felt dry and had no UI feedback to let the user know that their action actually did the job, this PR adds a simple image that acts as a ripple that gets displayed and animated when the user double taps on the screen.

### **Videos:**
**Phone**

https://github.com/jarnedemeulemeester/findroid/assets/45257975/8e3e97d8-7d1d-44a1-bc9b-d2cda49dc903

<br />

**Tablet**

https://github.com/jarnedemeulemeester/findroid/assets/45257975/d83f12d5-4716-4ee9-85af-c37d29dfd179


jarnedemeulemeester:
Nice! A ripple animation definitely improves the UX and gives the user feedback.

I just merged a PR to include a center area to play / pause. 
So I'll be updating this to also include a ripple animation for that.

PR #409, opened 2023-06-10
feat(atv): add `LEANBACK_LAUNCHER` category

Yash-Garg:
Add `LEANBACK_LAUNCHER` category so that the builds can be directly shown in the apps panel on Android TV's.


jarnedemeulemeester:
The phone version of Findroid is not meant to be used on Android TV.
A separate Android TV version in the works (see `android-tv` branch).

The TV version will require at lot more work before it is usable tho.

PR #413, opened 2023-06-16
fix: spelling error in remove address dialog

Zunawe:
Not tested. Saw it while using the app and it looked like an easy fix.


jarnedemeulemeester:
Thanks for fixing the spelling mistake!

I see that hardcoded strings are still used for the dialog... 
I will update them to string resources so they can be translated.

PR #432, opened 2023-07-07
feat: requeue pending downloads on downloads preference change

### LINKED ISSUE #430: Downloads doesn't start / stop automatically on settings change
**Describe the bug**
Toggling any of the 'Downloads' settings doesn't effect the active download items.

If you download an item while 'mobile data' or 'roaming' was off and you tried to toggle it on doesn't affect the downloads - you'll need to re-download (cancel and download again) to make it start downloading the items.

If any of the settings are on and you toggle it off, it will not stop the downloads (which isn't a big deal because you can cancel from the notification) but it will still be nice if it will cancel it automatically.

**To Reproduce**
1. Toggle 'Downloads' options off.
2. Turn off WiFi on your device.
3. Try to download something.
4. Toggle 'Downloads' options on.
5. Items aren't starting to download.

**Expected behavior**
Pending items should start downloading if the setting has been toggled on.
Active downloads should stop if settings has been toggled off.

**Screenshots**
N/A

 - Device: Pixel 6
 - Android version: 13
 - App version: v0.12.0
 - Jellyfin version: 10.8.10

**Additional context**
N/A

Natanel-Shitrit:
resolves #430

---

This feature requeues all pending download items when changing the 'Downloads' preferences, which saves the user from needing to cancel all pending downloads and download them again manually.

This only addresses the downloading problem mentioned in #430, but as mentioned there, the case when a user turns off either of the 'Downloads' preferences, it's possible to cancel the downloads from the notification tab.

### DIFF
--- a/app/phone/src/main/java/dev/jdtech/jellyfin/fragments/SettingsFragment.kt
+++ b/app/phone/src/main/java/dev/jdtech/jellyfin/fragments/SettingsFragment.kt
@@ -8,6 +8,7 @@ import androidx.preference.Preference
 import androidx.preference.PreferenceFragmentCompat
 import dagger.hilt.android.AndroidEntryPoint
 import dev.jdtech.jellyfin.AppPreferences
+import dev.jdtech.jellyfin.utils.Downloader
 import dev.jdtech.jellyfin.utils.restart
 import javax.inject.Inject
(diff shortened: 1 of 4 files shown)


jarnedemeulemeester:
No longer relevant since the compose rework

PR #433, opened 2023-07-07
fix: make external media follow app preferences

### LINKED ISSUE #421: Offline mode doesn't download external subtitles
**Describe the bug**
Offline mode doesn't seem to download external subtitles, rendering it somewhat useless for videos without built-in ones

**To Reproduce**
1. Get a video that has subtitles as an external file
2. Download it for offline playback
3. Start playback, tap on CC button, it does nothing if mpv player is not enabled, with mpv player it shows empty list; no subtitles are displayed during playback either way
4. After one deletes the offline download for the movie, playback now works with subtitles again

**Expected behavior**
Offline mode should also contain subtitles for the movie

 - Device: Samsung SM-P615
 - Android version: 13
 - App version: 0.10.1
 - Jellyfin version: 10.8.10

Natanel-Shitrit:
fixes #421

---

This change makes all external media downloads use the app preferences for whether or not to download over mobile data / roaming.


jarnedemeulemeester:
Thanks!

PR #439, opened 2023-07-12
Organize code, fix "locking" problem (#235)

### LINKED ISSUE #235: Login screen's back button doesn't do anything once server has been chosen
**Describe the bug**
On first start of Findroid, you get to choose a server, and then you move onto the login screen. However, if you don't actually log in, and close Findroid (swiping out of the app switch screen) the back button on the login screen doesn't do anything. To re-select a server, the app data must be cleared.

Note that the back button does work properly if you **don't** close the app while on the login screen.

**To Reproduce**
1. Clear Findroid app data
2. Open Findroid
3. Choose a server to proceed to the login screen
4. Open app switcher and swipe away Findroid
5. Reopen Findroid
6. Observe that back button does nothing when pressed

**Expected behavior**
Expected behavior is that the back button works at all times to permit reselecting the server.

**Device info:**
 - Device: Samsung Galaxy S21
 - Android version: 12
 - App version: 0.8.0
 - Jellyfin version: 10.8.8

### LINKED ISSUE #414: App "locks" itself when url changes.
**Describe the bug**
If the url you use for the server changes for any reason you cannot change the server from the login screen.

**To Reproduce**
1. Setup one url in findroid (eg. jelly.example.com)
2. Change/delete that url (eg. jellyfin.example.com)
3. Try to open findroid.

**Expected behavior**
Ability to change the server from the login screen.

 - Device: Samsung S20FE
 - Android version: 13
 - App version: v0.12.0
 - Jellyfin version: 10.8.10

Natanel-Shitrit:
# Related Issues
- Fixes #235 
- Fixes #414

---

Organize `MainActivity.kt`.

Open `serverSelectFragment` instead of `loginFragment` which could've lead to app locking if no user has logged in after adding the server.


jarnedemeulemeester:
Thanks a lot! I just moved the logic to the viewmodel instead of the fragment and added the necessary jellyfinApi update to use the new server url for logging in.

PR #443, opened 2023-07-19
refactor: use spacings to provide paddings

6xingyv:
To standardize the paddings, I put all common padding dp in Spacings() class and MaterialTheme.spacings


jarnedemeulemeester:
I didn't know you could do that with compose. I'm still pretty noob at it.

Anyway I updated the PR with the latest changes from the android-tv branch and fixed some linting issues.

This afternoon I will update the new screens with the spacings.

PR #450, opened 2023-07-28
feat: Add seek and play/pause functionality to onDoubleTap

djchittoor:
This patch enhances the `onDoubleTap` method within the media player component. It introduces a split-screen layout where the player view is divided into three equal areas (in the ratio of 2:1:2).

1. Leftmost Area: When double-tapped, it seeks the media playback backward by the defined seek increment
(`appPreferences.playerSeekBackIncrement`).

2. Middle Area: A double-tap in this area toggles the play/pause state of the player. If the player is currently playing, it will be paused, and if it's paused, it will be resumed.

3. Rightmost Area: When double-tapped, it seeks the media playback forward by the specified seek increment
(`appPreferences.playerSeekForwardIncrement`).

### DIFF
--- a/app/phone/src/main/java/dev/jdtech/jellyfin/utils/PlayerGestureHelper.kt
+++ b/app/phone/src/main/java/dev/jdtech/jellyfin/utils/PlayerGestureHelper.kt
@@ -68,13 +68,33 @@ class PlayerGestureHelper(
             override fun onDoubleTap(e: MotionEvent): Boolean {
                 // Disables double tap gestures if view is locked
                 if (isControlsLocked) return false
-                val viewCenterX = playerView.measuredWidth / 2
+
+                val viewWidth = playerView.measuredWidth
+                val areaWidth = viewWidth / 5 // Divide the view into 5 parts: 2:1:2
+
                 val currentPos = playerView.player?.currentPosition ?: 0
 
-                if (e.x.toInt() > viewCenterX) {
-                    playerView.player?.seekTo(currentPos + appPreferences.playerSeekForwardIncrement)
-                } else {
-                    playerView.player?.seekTo((currentPos - appPreferences.playerSeekBackIncrement).coerceAtLeast(0))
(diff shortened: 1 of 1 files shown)


jarnedemeulemeester:
Nice work, thanks!

MudssirAhmed:
`    
private fun togglePlayback() {
      playerView.player?.playWhenReady = !playerView.player?.playWhenReady!!
      animateRipple(activity.binding.imagePlaybackAnimationRipple)
  }
`

I think the togglePlayback() function is not very useful because the user will only play/pause the video using a double tap when the user is in plain mode(No Controls Mode) as you see in my Video-Recording.
As you added togglePlayback() for two features
1: Play -> Pause
2: Pause -> Play

So I think we have to add toggle functionality for both In-Plain-Mode as well as in Controls Mode.

[Video Recording Url](https://drive.google.com/file/d/1SfdKgqsDY2K44XBJTecpOAuF7Thn5LU9/view?usp=sharing)

Natanel-Shitrit:
> ` private fun togglePlayback() { playerView.player?.playWhenReady = !playerView.player?.playWhenReady!! animateRipple(activity.binding.imagePlaybackAnimationRipple) }`
> 
> I think the togglePlayback() function is not very useful because the user will only play/pause the video using a double tap when the user is in plain mode(No Controls Mode) as you see in my Video-Recording. As you added togglePlayback() for two features 1: Play -> Pause 2: Pause -> Play
> 
> So I think we have to add toggle functionality for both In-Plain-Mode as well as in Controls Mode.
> 
> [Video Recording Url](https://drive.google.com/file/d/1SfdKgqsDY2K44XBJTecpOAuF7Thn5LU9/view?usp=sharing)

I think we need a broader solution for this problem, not a fix for this *specific* instance.

I propose to allow the user to configure gestures as they like from the settings (with a default layout)

This will probably be a-bit of work, but I think it will be a worthwhile solution.

PR #465, opened 2023-08-10
Fix playback sync issue

### LINKED ISSUE #464: Playback progress too far after exiting offline mode
**Describe the bug**
When partially watching a video in offline mode and turning it off, the playback progress is synced to the server. After a few minutes, the playback progress on the server is ahead of where you actually stoppe watching. When checking the activity in jellyfin just after syncing, it shows the episode as currently playing, so I suspect findroid reports the episode as currently playing instead of having ended playback of the episode.

**To Reproduce**
1. Download a video
2. Enable offline mode
3. Watch the video halfway
4. Disable offline mode
5. Wait a few minutes
6. Jellyfin now reports the video as being played more than halfway

**Expected behavior**
Jellyfin reports the video being played to where you acually played it

Jcuhfehl:
Fixes #464 by using `OnPlaybackStopped` instead lf `OnPlaybackProgress` for syncing.


jarnedemeulemeester:
Thanks!

PR #466, opened 2023-08-12
feat: Chapter additions

### LINKED ISSUE #158: [Feature Request] Video Chapter Features
If there's chapter information in the video, I'd like to be able to move the chapter.
also, If the name of the chapter is OP, Opening, 여는곡, 오프닝, オープニング, 開く曲 etc., I would like to have an opening skip button.

### LINKED ISSUE #437: [Feature Request] Support for MKV chapters
Ideas for implementation (highly recommend 3 & 4)

1) A menu like VLC. (not recommended)

2) Tap the title for a a drop-down menu.

3) New previous/next buttons next to the "lock" button.

4) Seek bar gets split into chapters with the chapter title above + haptic feedback and magnetic pull to the start of each chapter. (Kinda like YouTube)

5) Setting toggle to navigate chapters with the previous/next episode buttons + long press to navigate episodes.

6) Simple circles on the seek bar to indicate chapters like (MPC)

Natanel-Shitrit:
### PR Additions:
- Chapter markers.
markers
- Move to previous / next chapter on left / right long press on screen.
skip-chapter
- Chapter name on skip.
chapter-name
- App preferences.
Screenshot_20230812-155849~3
Screenshot_20230812-155849~2

## Issues
- Resolves #158
- Resolves #437 
## TODO:
1. ~~Implement chapter extraction from ExoPlayer. ([StackOverflow](https://stackoverflow.com/questions/65659175/exoplayer-get-chapters-of-file))~~
~~Dropped ExoPlayer because I couldn't find any way to get the metadata 😢~~
ExoPlayer now supported, thanks @jarnedemeulemeester for pointing out Jellyfin API chapter info!
2. ~~Implement app preference for turning off / on markers / gesture.~~
Done.
3. ~~Use system "Material You" theme for markers.~~
This is probably fine as white (timebar contrast color)
4. ~~Fix code reuse in `PlayerGestureHelper.kt`.~~
Would be better to refactor in a new PR with other changes to the same file.
5. ~~Add chapter name in player title.~~
I think I'll keep it simple without changing the title for now.

## Credits:
- [\[StackOverflow\] How to add markers in ExoPlayer](https://stackoverflow.com/questions/49314638/how-to-show-ad-markers-in-exoplayer)
- [\[GitHub - mpv-android\] How to get chapter times in MPV player](https://github.com/mpv-android/mpv-android/blob/master/app/src/main/java/is/xyz/mpv/MPVView.kt#L274-L287)


Natanel-Shitrit:
@jarnedemeulemeester I think this is ready 😄
If you have any suggestion, would be happy to implement!

Natanel-Shitrit:
Rebased with main PiP changes

Natanel-Shitrit:
First version crashed when skipping an episode, because it tried to load the chapters too early.
This is now fixed and ready for review 😄
@jarnedemeulemeester

garvongithub:
Thanks for doing this. I don't know how helpful it can be but there is this [player](https://github.com/moneytoo/Player) which also uses exoplayer and has chapter support.

Natanel-Shitrit:
> Thanks for doing this. I don't know how helpful it can be but there is this [player](https://github.com/moneytoo/Player) which also uses exoplayer and has chapter support. 

That's certainly better than anything.
I'll look into this :)
Thanks!

Natanel-Shitrit:
Seems like a lot of work...
https://github.com/moneytoo/Player/commit/63c920076f454ba5707055148756a71e19552f66

He pulls the chapters from the media info he gets from FFmpeg library.
I originally tried this and gave up because it felt too much work for the insignificant importance of such feature.

I think the MPV player is mature enough to be preferred over ExoPlayer and thus it's not really a big deal.
But, of course anyone is welcome to try and implement this for the ExoPlayer.

Natanel-Shitrit:
Oops, I didn't mean to merge this.
Please don't merge until I fix this.

jarnedemeulemeester (changes requested):
Hi @Natanel-Shitrit 
Thanks a lot for implementing this!

I do have a few suggestions.
I see you are using mpv to get the chapters from the item and that's the reason why it is only available when using mpv. But apparently Jellyfin already extracts the chapters from items as well.
You should be able to use the API to retrieve the chapters and display them in both Exoplayer and mpv :smile: 

It seems to be part of the `BaseItem` so no need to call an extra API endpoint. You probably only need to add a few properties to some `Findroid*` classes and maybe create a new `FindroidChapter` class or something like that.
```
"Chapters": [
    {
        "StartPositionTicks": 0,
        "Name": "string",
        "ImagePath": "string",
        "ImageDateModified": "2019-08-24T14:15:22Z",
        "ImageTag": "string"
    },
]
```
I think we can ignore the Image part of the chapter for now since this has to be manually enabled for each library anyway.

Also I recently merged a PR which add 2x playback speed on long press. That obviously conflicts with the skipping to next chapter on long press. So we have to change one or the other, but I don't really know which one yet and to what action.

Jcuhfehl:
> Also I recently merged a PR which add 2x playback speed on long press. That obviously conflicts with the skipping to next chapter on long press. So we have to change one or the other, but I don't really know which one yet and to what action.

I already mentioned this in the discord, but I think using the next/previous episode buttons for chapters would be a good option, since it feels like it's doing a similar thing. Another option might be the long-press on the fast forward button.

I think long pressing the right of the screen is a very appropriate gesture for 2x so I'd prefer leaving that unchanged

Natanel-Shitrit:
@jarnedemeulemeester Thanks for the detailed response!

That's a nice surprise that Jellyfin already extracts the chapter info, this will make it easier for comparability since it will not rely on the MPV functionality.

I'll try and Implement a POC for the chapter extraction from Jellyfin.

Regarding the skip chapter gesture I think giving the user the ability to customize it will be the best approach here - obviously this will take time and effort but I think the best way to implement this is to create a settings page for player gestures - list of default gestures, and give the ability to remove / modify / add from that list.

Each gestures can be represented like this:
- Tap Type - Double tap / Long press
- Location - Right / Middle / Left / All
- Action - Action to perform

I'll open a feature request and try to implement this too.

But probably only after this will be merged since it will probably something bigger.

Natanel-Shitrit:
@garvongithub @jarnedemeulemeester Now both players are supported!
I've successfully migrated to Jellyfin API for getting information about chapters.

Natanel-Shitrit:
# TODO:
- Integrate chapter skipping in the "skip media" buttons.
> when chapters are enabled and available for specific media, clicking on previous / next media buttons in the player will skip chapters instead of the media it-self.
- Save chapter info for local items (downloaded media) to support new features offline.

Natanel-Shitrit:
Thanks for testing!
 
Feel free to fix / change anything you see fit!

if you want - I can tackle the bugs on the weekend but I guess you can do that much faster and better than me😅

jarnedemeulemeester:
- Fixed seeking in mpv by returning the chapter that is being sought to
- Simplified `getCurrentChapterIndex` 
- Seeking to previous chapter now returns the current chapter if playback progress is more than 5 seconds past the start of the chapter
- Added some docstring
- Changed the naming to just "Chapters"

Still need to fix the database issue 😅

jarnedemeulemeester:
@Natanel-Shitrit Yep, that fixed it!

jarnedemeulemeester:
I will do one final check of everything. If nothing goes wrong I will merge it!

jarnedemeulemeester (approved):

Natanel-Shitrit:
I think the only thing we forgot is to add the new feature in the `README.md`, I'll open a new PR 😄

PR #468, opened 2023-08-14
fix episode indicator wrapping on second line

### LINKED ISSUE #445: Three-digit numbers in the unwatched episode indicators are cut off.
**Describe the bug**
If a series has 100 or more unwatched episodes, the number wraps and is cut off.

**To Reproduce**
1. Have any series with 100 or more unwatched episodes. 

**Expected behavior**
The number should display in one line with no cropping.

**Screenshots** (See Bleach in the screenshot)
Screenshot_20230722_073431_Findroid

 - Device: Samsung Galaxy Note 20 Ultra 5G
 - Android version: 13
 - App version: v0.12.0
 - Jellyfin version: 10.8.8

teobaranga:
Fixes #445

The episode indicator background was constrained to a 24dp box. If this box could not contain the count, then the count would wrap onto a new line.

The proposed solution is to use a rounded rectangle as the background instead. This can then expand into a chip as needed to accommodate larger numbers without wrapping.

See below how this looks like with 1, 2, and 4 digits:

### DIFF
--- a/app/phone/src/main/res/layout/base_item.xml
+++ b/app/phone/src/main/res/layout/base_item.xml
@@ -59,11 +59,13 @@
 
         <TextView
             android:id="@+id/item_count"
-            android:layout_width="24dp"
+            android:layout_width="wrap_content"
             android:layout_height="24dp"
             android:layout_marginStart="8dp"
-            android:background="@drawable/circle_background"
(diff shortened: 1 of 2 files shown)


jarnedemeulemeester:
Thanks a lot, looks good!

Issue #384, opened 2023-05-27
Feature Request - Vertical scrolling pagination in collections screen + ability to sort collection by release date


MakD:
![Screenshot_20230527-153245.png](https://github.com/jarnedemeulemeester/findroid/assets/8217200/8b175f11-968a-491b-8e66-97ac5ecc7786)

Currently the scrolling is horizontal, which leads to a lot of wasted real estate in big movie collections, and also, sorting by release date isn't possible

Natanel-Shitrit:
Hey!

Thanks for reporting 😄
Regarding the vertical scrolling pagination, this is already mentioned in an earlier issue and it's tracked in #94.

Please make sure to check if there is an open issue regarding your problem before opening a new one!

Regarding the other request - please open one issue per one bug / request - this helps organize things better!

MakD:
Hello @Natanel-Shitrit any update with this. Is it in plans for implementation?

jarnedemeulemeester:
Closing this since scrolling is now vertical. Extra sorting methods will come later

Issue #94, opened 2022-02-18
[Feature request] favorites overview


BluePixel4k:
A lot of space is wasted in the favorites overview if you only have movies marked. 
An automatic switch between horizontal and vertical scrolling would be great.

![Screenshot_20220218-164841](https://user-images.githubusercontent.com/77979781/154715765-65eca577-6a33-4b89-b473-6bcb1f10d9bf.png)

jarnedemeulemeester:
Now scrolls vertically in the compose rework

PR #475, opened 2023-08-17
set sort order for collection view to release date

arnokeesman:
Until there's an option to set the order in collection views I believe the default should be by RELEASE_DATE, other FJ interfaces do it like this as well and it just makes the most sense.

Related issues:
#384


Natanel-Shitrit:
+ #94

Unkorneglosk:
This would makes way more sense actually

arnokeesman:
implemented in 1.14.1

PR #476, opened 2023-08-18
fix: snackbar not showing in landscape mode

### LINKED ISSUE #470: Feature request  Offline mode toggle in the downloads screen
Would it be possible to put a toggle for "Offline Mode" in the downloads screen.  Both of the tablets i use do not show the "switch to offline mode" popup (for lack of a better word) and going into the settings is a couple steps too many for my kids.

Maybe better yet, could the downloads screen always behave as if offline mode is toggled? no need to show the splash screen for the movie/show there, just the ability to play the media.

Natanel-Shitrit:
This affected also all tablets since they use the landscape view!

Added `setTextMaxLines` because for some reason the text is truncated on tablets.

- Fixes #470.

Thanks @Jcuhfehl for the help :D


jarnedemeulemeester:
Thanks!

PR #478, opened 2023-08-19
feat: add tagline for movies

### LINKED ISSUE #451: [feature request] add taglines for films.
Almost all films have taglines (as seen here in the matadata and in the jellyfin app)
Screenshot_20230728-184709
Screenshot_20230728-184721
But it's not showing in findroid?
Screenshot_20230728-184728
Is it a possibility?

Natanel-Shitrit:
## Feature Preview
screen

## Related Issues
- Resolves #451


jarnedemeulemeester:
I am not the biggest fan of a tagline (especially on a space constrained mobile screen) but if we do decide to add it I would put it underneath the buttons I think

Natanel-Shitrit:
> I am not the biggest fan of a tagline (especially on a space constrained mobile screen) but if we do decide to add it I would put it underneath the buttons I think.
> 
> @Natanel-Shitrit could you try that and see if it looks good?

Sure, I'll move it later today and test :)

Natanel-Shitrit:
> I am not the biggest fan of a tagline (especially on a space constrained mobile screen) but if we do decide to add it I would put it underneath the buttons I think.
> 
> @Natanel-Shitrit could you try that and see if it looks good?



Looks good

Natanel-Shitrit:
Looks good

I would actually make the text bigger, what do you think?

Natanel-Shitrit:
Looks like this now.
screen

Natanel-Shitrit:
What do you think about moving it to the page app bar? (top bar where you have the name of the movie)
It's currently just a duplicating the movie name so it could be nice 😄

Natanel-Shitrit:
I guess we can close this after deciding it's not really necessary.
Up to you @jarnedemeulemeester 😄

PR #482, opened 2023-08-21
Add support for TV Theme Songs

### LINKED ISSUE #105: [Feature Request] Support for TV Theme Songs
The app to add support for playback of theme songs when a TV Show/episode is selected - just like how Jellyfin does it.

teobaranga:
Closes #105 

Adds a new toggle in settings that controls whether theme songs play automatically when browsing TV shows. Playback is started only when navigating to the show page. This continues until either the app goes to the background, the user navigates back, or streaming starts. The audio fades in and out to smooth out songs that start loudly.

The location of the new classes may not be ideal but I wanted to minimize the addition of new dependencies to existing modules or creating new modules. I can revisit that if needed.

### Demo
Turn up the audio if needed, the recording has weak audio for some reason.

https://github.com/jarnedemeulemeester/findroid/assets/139957/092f0d73-b959-41da-a032-1a753031d49f


jarnedemeulemeester:
I don't think I want this feature in Findroid. 
In my opinion it is just annoying to have music playing while browsing in your library.
If it was just one line of code then sure, but it's quite a lot of code to have this working.

PR #518, opened 2023-09-19
feat: sleep mode

AstroTheAstronaut:
This PR adds a **Sleep Mode** to Findroid.  (Fixes a request on Discord)
At the moment, if the user falls asleep while watching a show, the screen will remain on for the remainder of the show/movie, maybe even more, depending if it's a show or a movie (if it is a show, the rest of the episodes in the series will play). This isn't ideal as it consumes battery needlessly. 
This PR implements a timer that, when it's finished, shows a new screen that explains that Sleep Mode has been enabled and that the screen will turn off on its own to prevent unnecessary battery consumption.
This feature can be disabled in the app's settings if people don't want to have this option (Not sure if this should be removed)

**Screenshots:**

Screenshot_20230919-122416
_Setting in the Player section_

Screenshot_20230919-122524
_Button next to the lock button to show the Sleep Mode button_

Screenshot_20230919-122531
_Sleep Mode timer dialog_

Screenshot_20230919-122650
_Sleep Mode screen with unlock button_

**To do:**

- [ ] Test for bugs
- [ ] Add option for "Until media ends" in the dialog
- [ ] Test on big screens

**UPDATE:** Tested it with MPV and it works just fine also.

### DIFF
--- a/app/phone/src/main/java/dev/jdtech/jellyfin/PlayerActivity.kt
+++ b/app/phone/src/main/java/dev/jdtech/jellyfin/PlayerActivity.kt
@@ -438,7 +438,8 @@ class PlayerActivity : BasePlayerActivity() {
     }
 
     private fun showSleepTimerDurationDialog() {
-        val timerTexts = listOf("1 minute", "5 minutes", "10 minutes", "30 minutes", "1 hour", "2 hours")
+        val timerTexts =
+            listOf("1 minute", "5 minutes", "10 minutes", "30 minutes", "1 hour", "2 hours")
         val durationValues = longArrayOf(60000L, 300000L, 600000L, 1800000L, 3600000L, 7200000L)
         var selectedDuration = durationValues[2] // Default: 10 minutes in milliseconds
 
@@ -461,8 +462,7 @@ class PlayerActivity : BasePlayerActivity() {
                 dialog.dismiss()
             }
             .setCancelable(true)
-            wasDialogShown = true
-
+        wasDialogShown = true
         builder.create().show()
     }
 
--- a/app/phone/src/main/java/dev/jdtech/jellyfin/PlayerActivity.kt
+++ b/app/phone/src/main/java/dev/jdtech/jellyfin/PlayerActivity.kt
@@ -174,7 +174,7 @@ class PlayerActivity : BasePlayerActivity() {
                                 speedButton.imageAlpha = 255
                                 pipButton.isEnabled = true
                                 pipButton.imageAlpha = 255
-								sleepModeButton.isEnabled = appPreferences.sleepMode
+                                sleepModeButton.isEnabled = appPreferences.sleepMode
                                 sleepModeButton.imageAlpha = if (appPreferences.sleepMode) {
                                     255
                                 } else {
--- a/app/phone/src/main/java/dev/jdtech/jellyfin/PlayerActivity.kt
+++ b/app/phone/src/main/java/dev/jdtech/jellyfin/PlayerActivity.kt
@@ -66,7 +66,7 @@ class PlayerActivity : BasePlayerActivity() {
     override val viewModel: PlayerActivityViewModel by viewModels()
     private var previewScrubListener: PreviewScrubListener? = null
     private var sleepJob: Job? = null
-    private var wasDialogShown : Boolean = false
+    private var wasDialogShown: Boolean = false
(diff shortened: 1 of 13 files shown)


app/phone/src/main/java/dev/jdtech/jellyfin/PlayerActivity.kt
```diff
@@ -382,4 +420,72 @@ class PlayerActivity : BasePlayerActivity() {
...
+                dialog.dismiss()
+            }
+            .setCancelable(true)
+        wasDialogShown = true
+        builder.create().show()
+    }
+
+    private fun disableSleepMode() {
+        binding.playerView.player?.playWhenReady = true
+        window.addFlags (WindowManager.LayoutParams.FLAG_KEEP_SCREEN_ON)
```
Natanel-Shitrit, line 468: Lint issue here, just remove the space after `addFlag`.

Natanel-Shitrit (changes requested):

app/phone/src/main/java/dev/jdtech/jellyfin/PlayerActivity.kt
AstroTheAstronaut: Thank you!

Natanel-Shitrit (approved):

app/phone/src/main/java/dev/jdtech/jellyfin/PlayerActivity.kt
```diff
@@ -163,6 +174,13 @@ class PlayerActivity : BasePlayerActivity() {
                                 speedButton.imageAlpha = 255
                                 pipButton.isEnabled = true
                                 pipButton.imageAlpha = 255
+                                if (appPreferences.sleepMode) {
+                                    sleepModeButton.isEnabled = true
+                                    sleepModeButton.imageAlpha = 255
+                                } else {
+                                    sleepModeButton.isEnabled = false
+                                    sleepModeButton.imageAlpha = 0
+                                }
```
Yash-Garg, line 183: Can do this instead here.

```suggestion
								sleepModeButton.isEnabled = appPreferences.sleepMode
                                sleepModeButton.imageAlpha = if (appPreferences.sleepMode) {
                                    255
                                } else {
                                    0
                                }
```

app/phone/src/main/java/dev/jdtech/jellyfin/PlayerActivity.kt
```diff
@@ -382,4 +420,72 @@ class PlayerActivity : BasePlayerActivity() {
...
+        val hourText = if (hours == 1L) "hour" else "hours"
+        val minuteText = if (minutes == 1L) "minute" else "minutes"
+
+        return when {
+            hours > 0 && minutes > 0 ->
+                "Sleep Mode will be enabled in $hours $hourText and $minutes $minuteText"
+            hours > 0 ->
+                "Sleep Mode will be enabled in $hours $hourText"
+            else ->
+                "Sleep Mode will be enabled in $minutes $minuteText"
```
Yash-Garg, line 488: Moving this as a string resource will be better for Translations.

AstroTheAstronaut: > Moving this to a string resource will be better for Translations.

I think I moved everything to a string resource, but I had to use @SuppressLint("StringFormatMatches") for that function because I couldn't figure out what the thing wanted from me. It works fine tho.

Yash-Garg: @AstroTheAstronaut `StringFormatMatches` lint is probably because you used `%1$s` while you are passing a number value which expects `%1$d` in some places like `hours`, `minutes`, etc.

arjunluthra:
Poking this feature, hope this can get in soon.  Looking forward to it!

joshjryan:
I realize that this PR has been open for over a year, but for now some reason I I was just thinking about it.

I wouldn't like to propose/request that this feature (and the UI elements for it) be named "Sleep Timer" (instead of Sleep Mode). In my experience, that is usually what this type of feature is called on TVs and related apps.

aslafy-z:
Hello, I'm interested with this feature and would love to see it land into Findroid. The UX looks good to me. Aligned with @joshjryan for the naming. Do you need help to make this PR land? I can install the flutter SDK, rebase it and try it locally if you want.

PR #529, opened 2023-09-28
refactor: remove `cleanUpOldDownloads`

Natanel-Shitrit:
Pretty sure enough time has passed since this was introduced and we can remove it :)

### DIFF
--- a/app/phone/src/main/java/dev/jdtech/jellyfin/MainActivity.kt
+++ b/app/phone/src/main/java/dev/jdtech/jellyfin/MainActivity.kt
@@ -1,11 +1,9 @@
 package dev.jdtech.jellyfin
 
 import android.os.Bundle
-import android.os.Environment
 import android.view.View
 import androidx.activity.viewModels
 import androidx.appcompat.app.AppCompatActivity
-import androidx.lifecycle.lifecycleScope
 import androidx.navigation.NavController
 import androidx.navigation.NavGraph
 import androidx.navigation.fragment.NavHostFragment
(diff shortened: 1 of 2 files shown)


jarnedemeulemeester (changes requested):
I agree, we should remove it now. There are only a handfull of people still on an older version before the downloads rework. And those people probably cannot update because their device is running an old Android version.

I think we should also delete the `appPreference` `downloadsMigrated` variable :smile: 
https://github.com/jarnedemeulemeester/findroid/blob/d65591a19ec3beefc7daaf63ec6811777a41d8d7/preferences/src/main/java/dev/jdtech/jellyfin/AppPreferences.kt#L148C1-L153

Natanel-Shitrit:
> I agree, we should remove it now. There are only a handfull of people still on an older version before the downloads rework. And those people probably cannot update because their device is running an old Android version.
> 
> I think we should also delete the `appPreference` `downloadsMigrated` variable :smile: 
> https://github.com/jarnedemeulemeester/findroid/blob/d65591a19ec3beefc7daaf63ec6811777a41d8d7/preferences/src/main/java/dev/jdtech/jellyfin/AppPreferences.kt#L148C1-L153

Oops, I totally forgot abot that, thanks - I'll add it once I'm home!

jarnedemeulemeester (approved):
I rebased the PR onto main and removed the `downloadsMigrated` prefrence since it was a small change I could easily commit

Natanel-Shitrit:
> I rebased the PR onto main and removed the `downloadsMigrated` prefrence since it was a small change I could easily commit

Thank you 🙏
I'm sorry I wasn't able to complete this PR.

Hard times - hope to return supporting Findroid soon ♥️

jarnedemeulemeester:
> Thank you 🙏 I'm sorry I wasn't able to complete this PR.
> 
> Hard times - hope to return supporting Findroid soon ♥️

No problem man! I haven't been able to do much myself either. 

Take care ♥️

PR #266, opened 2023-02-02
Added a preferred stream quality option

e2fo2l:
Added the option to select a preferred stream quality in the player settings. This works for both ExoPlayer and mpv.
You can either request the original file, or ask for a specific resolution (4K, 1080p, 720p, 480p, 360p).
If the asked resolution is lower than the original file's, it will require transcoding.

Ideally, the option to switch the quality should have been directly inside the player, however, I haven't found an easy way to set different stream URLs for different resolutions.

Please note that this option is only for video streaming, downloads still use the original file regardless.

PR #555, opened 2023-10-22
Transcoding Support

seang96:
This is a PR with the changes from #266 merged with latest code after the rework.

From testing, the only issues I have encountered so far are:

1. audio tracks options do not load, default is the only one that can be selected
2. Website says it is direct playing, but it does transcode, I don't think the original PR is sending all of the correct API requests to make Jellyfin say it is being transcoded. I believe if we se t that maximum bit rate to a low value it may fix this, but I think we'd need to set predefined bitrates per resolution?
3. No option to change transcoding in the player


peerless2012:
This feature is useful, will update this pull request?

nomadics9:
jellyfinApi.api.dynamicHlsApi.getVariantHlsVideoPlaylistUrl()

doesnt work. playback shows nothing

jellyfinApi.videosApi.getVideoStreamUrl()

works with transcoding but playback is unseekable and shows as livestream rather than the full playback with the full media time.

idk why and i dont know how to fix it :(

Issue #495, opened 2023-08-27
[Feature request] Access season from episode


Margit420:
ie tap "S3:E8 - Title" to access Season 3

![Screenshot_2023-08-27-14-13-43-09_aff99014c254344a005aaf8c1683039d](https://github.com/jarnedemeulemeester/findroid/assets/113560376/7794a960-1957-4da0-b9e4-fab82c9fbfc6)

jarnedemeulemeester:
You can now access the season screen from the episode screen in the compose rework

PR #563, opened 2023-11-06
open season view from episode

arnokeesman:
clicking episode_name navigates to the season for that episode

seasonName added to episode classes so the opened season view can have it's name

#495


Julien9969 (approved):

jarnedemeulemeester:
Has been added in the compose rework

PR #622, opened 2024-01-11
chore: use type-safe project accessors

Yash-Garg:
Use type-safe project accessors ([introduced in Gradle 7](https://docs.gradle.org/7.0/userguide/declaring_dependencies.html#sec:type-safe-project-accessors))


jarnedemeulemeester:
I'm not a huge fan of using experimental features.
But I like this so I'll merge it :smile:

jarnedemeulemeester (approved):

Issue #224, opened 2022-12-22
Add support for Mixed Folders of movies and series.


JulioNahum:
**Describe the bug**
When creating a mixed library (includes movies and series) on the jellyfin server, Findroid doesn't detect them correctly, and the application displays it out of order. The chapters come loose, the seasons on the other hand.

**Screenshots**

https://user-images.githubusercontent.com/30002132/209163785-2c59eda8-08e6-475e-b38b-9890a85d4493.mp4

https://user-images.githubusercontent.com/30002132/209163809-a679ea4b-8e76-4ff3-818b-f1b840cfd720.mp4

**Comparing with the official Jellyfin app for android.**

https://user-images.githubusercontent.com/30002132/209164962-9c839e4d-82b5-4ca8-bcdc-dbbf6f3ace31.mp4

My library is very well scanned, therefore in the official Jellyfin apps it shows them very well ordered. I love Findroid. That's why I want you to add this support or fix this bug.

**Device info (please complete the following information):**
 - Device: [Motorola Edge 20 PRO]
 - Android version: [12]
 - App version: [v0.8.0 (14)]
 - Jellyfin version: [10.8.8]

**Additional context**
On the main screen the content is displayed correctly. But when entering the libraries and reviewing the content it has, the bug of showing absolutely everything starts in a very messy and stressful way.

Issue #497, opened 2023-08-28
Findroid app can't show media libraries of type "Mixed Movies and Shows"


xenoncolt:
**Describe the bug**
This is the current version app view where not show "Mixed Movies and Shows"
![WhatsApp Image 2023-08-28 at 17 36 03](https://github.com/jarnedemeulemeester/findroid/assets/81591350/a766c87c-4f85-4405-8817-ff7bb00b28d9)

Here I marked those Movies and shows that are not displayed
![image](https://github.com/jarnedemeulemeester/findroid/assets/81591350/a97f1306-ffd2-49a2-94d1-d71a838bf311)

I hope you understand what I want to say.

gjp1120:
same issue

UriSheinman:
Not really a bug, just a problem with how findroid looks for libraries. It only looks for ones marked as Movies/ Shows and the collections folder. Thats why music won't appear and why mixed folders won't appear. It just doesn't look for them.

GreatestFool:
True enough, though an improperly scoped domain doesn't invalidate the report either. It may not be a bug, but it is a feature that is wanted and sorely lacking. The music and book libraries I can understand, but given that a mixed library is generally used for both TV-shows and Movies, I think it has its place in this application. It's already pretty great, after all.

I do hope we see it implemented someday soon. Though, giving the view models a quick one-over, I'd guess they'd need at least some changes to avoid code-reuse in a model dedicated for mixed libraries. I'd like to say that shouldn't be too hard to implement, but I'm not familiar enough with the code-base to really say anything there.

GreatestFool:
For sake of keeping things _somewhat_ organized, here's other duplicate issues:
- https://github.com/jarnedemeulemeester/findroid/issues/154
- https://github.com/jarnedemeulemeester/findroid/issues/224
- https://github.com/jarnedemeulemeester/findroid/issues/352

Feel free to ignore this message, this is mostly so that these issues are shown to be referenced in this one.

Issue #352, opened 2023-05-10
Findroid app can't show media libraries of type "Mixed Movies and Shows"


maulik-sharma:
**Describe the bug**
Findroid app can't show media libraries of type **Mixed Movies and Shows**

**To Reproduce**
While creating a library in jellyfin set a library's conent type to **Mixed Movies and Shows** will caues it to not display that library in the Findroid app.

**Expected behavior**
Library of content type Mixed Movies and Shows should also be displayed among other libraries in findroid app

**Device info (please complete the following information):**
 - Device: Xiaomi 11i
 - Android version: 13
 - App version: 0.11.0
 - Jellyfin version: 10.8.10

**Additional context**
I have a jellyfin library of content type Mixed Movies and Shows which I use for anime. All the other libraries are properly displayed in FIndroid app but the anime library is not available

decisoft:
For the record, this is the traceback:

`java.lang.NullPointerException
 java.lang.Exception: java.lang.NullPointerException
	at c8.u1.invoke(Unknown Source:144)
	at k3.b1.b(Unknown Source:109)
	at k3.f.a(Unknown Source:31)
	at k3.i3.m(Unknown Source:689)
	at k3.i3.f(Unknown Source:12)
	at ja.q.e0(Unknown Source:4)
	at com.bumptech.glide.f.E0(Unknown Source:71)
	at k3.y.a(Unknown Source:2290)
	at pa.t.e(Unknown Source:4)
	at pa.s.o(Unknown Source:133)
	at pa.s.a(Unknown Source:0)
	at k3.k.a(Unknown Source:77)
	at kotlinx.coroutines.flow.u.a(Unknown Source:102)
	at kotlinx.coroutines.flow.w0.a(Unknown Source:2)
	at kotlinx.coroutines.flow.k0.l(Unknown Source:187)
	at kotlinx.coroutines.flow.j0.m(Unknown Source:12)
	at x9.a.h(Unknown Source:8)
	at kotlinx.coroutines.m0.run(Unknown Source:102)
	at kotlinx.coroutines.x0.o0(Unknown Source:38)
	at kotlinx.coroutines.l.o(Unknown Source:107)
	at oa.b.c(Unknown Source:6)
	at oa.m.p(Unknown Source:89)
	at oa.j.j(Unknown Source:0)
	at k3.t2.i(Unknown Source:31)
	at k3.t2.f(Unknown Source:505)
	at k3.c2.m(Unknown Source:11)
[...]

Issue #154, opened 2022-08-28
Support for Mixed Content Libraries (Movies and series)


JulioNahum:
**Describe the bug**
Right now when adding a library with mixed content, the chapters of the series are combined with the movies and vice versa.

**Screenshots**

https://user-images.githubusercontent.com/30002132/187086610-cfa76838-acfc-4439-a03a-35d43af14588.mp4

**Device info (please complete the following information):**
 - Device: [MOTOROLA EDGE 20 PRO]
 - Android version: [ANDROID 12]
 - App version: [v0.6.0]

**Additional context**
Someday there will be that support?****

AnonymousWebHacker:
Me parece un problema del servidor jellyfin, no de Findroid app. Comprueba si el contenido esta mixeado asi usando jellyfin cliente

PR #624, opened 2024-01-12
Add support for mixed libraries

Jcuhfehl:
So this turned out to be rather trivial.  I basically just added mixed libraries to the list of supported types and it worked.

I don't use mixed libraries, so I haven't thoroughly tested it, but everything seems to work fine, including downloads.

Closes: #224 #154 #352 #497


Jcuhfehl:
Superceded by #628

PR #628, opened 2024-01-20
feat: mixed libraries

nels0nwu:
Mixed libraries to show up in 'My Media' and 'Home' views
Shows mixed libraries as collections of Movies, TV shows and Folders you can navigate into. Modelled on what the official client does.

Closes: https://github.com/jarnedemeulemeester/findroid/issues/224
Closes: https://github.com/jarnedemeulemeester/findroid/issues/154
Closes: https://github.com/jarnedemeulemeester/findroid/issues/352
Closes: https://github.com/jarnedemeulemeester/findroid/issues/497
Closes: https://github.com/jarnedemeulemeester/findroid/issues/669


app/phone/src/main/java/dev/jdtech/jellyfin/fragments/LibraryFragment.kt
```diff
@@ -222,6 +224,15 @@ class LibraryFragment : Fragment() {
                     ),
                 )
             }
+            is FindroidFolder -> {
+                findNavController().navigate(
+                    LibraryFragmentDirections.actionLibraryFragmentSelf(
+                        item.id,
+                        item.name,
+                        CollectionType.Mixed,
```
nels0nwu, line 232: Would be better without this collectiontype assumption

Natanel-Shitrit:
Please add a separate "Closes #<issue>" for each issue - otherwise it will close only the first one

jarnedemeulemeester (approved):
Looks good!
Tested a mixed library with subfolders and all seems fine. 

Happy to see you also implemented the routing in the TV version :smile: 

The only thing I changed is adding some default values to `FindroidFolder` instead of retrieving them from the `BaseItem` since they don't seem to apply to folders.

PR #633, opened 2024-01-22
feat: skip credits

### LINKED ISSUE #620: Support on skipping credits in intro skipper
Credits not able to be skipped

Avaliable in [https://github.com/jumoog/intro-skipper]()'s build
Use"IntroSkipperSegments"(support both intro and credits) at first and use "IntroTimestamps"(this only support intro)
Format of IntroSkipperSegemets:
```
{
    "Introduction": {
        "EpisodeId": "1e3cf41f1ae051f68ab56980f937f745",
        "Valid": true,
        "IntroStart": 145.792,
        "IntroEnd": 228.497,
        "ShowSkipPromptAt": 140.792,
        "HideSkipPromptAt": 155.792
    },
    "Credits": {
        "EpisodeId": "1e3cf41f1ae051f68ab56980f937f745",
        "Valid": true,
        "IntroStart": 1313.264,
        "IntroEnd": 1408.2640000000001,
        "ShowSkipPromptAt": 1308.264,
        "HideSkipPromptAt": 1323.264
    }
}
```

**Additional context**
This should only introduce some small modification without breaking the original support, hope this can be implemented.

cd16b:
Support for jumoog's **Intro Skipper**:
- FindroidSegment
- Skip Intros
- Skip Credits

~Keep using the old API for intros, and still support ConfusedPolarBear's Intro Skipper.~

fix #620


Gauvino:
Does this could support https://github.com/endrl/jellyfin-plugin-media-analyzer ?

cd16b:
> Does this could support https://github.com/endrl/jellyfin-plugin-media-analyzer ?

This PR only allows skipping credits from Jumoog's Intro Skipper v0.1.8.0, a new release of ConfusedPolarBear's Intro Skipper.
The plugin "endrl/jellyfin-plugin-media-analyzer" needs the segments API, but it is not currently supported by Jellyfin version 10.8.xx. Therefore, we will have to wait at least for Jellyfin 10.9.0.

jumoog:
I only forked the plugin for myself because ConfusedPolarBear quit. I also left everything as it is so that you only have to change the repo url and everything continues to work.

AbandonedCart:
https://github.com/jarnedemeulemeester/findroid/blob/main/player/video/src/main/java/dev/jdtech/jellyfin/viewmodels/PlayerActivityViewModel.kt#L253

Seems silly to submit an entire PR for this, but you want the line above to check the hideSkipPromptAt value - 1.

cd16b:
> Seems silly to submit an entire PR for this, but you want the line above to check the hideSkipPromptAt value - 1. Thought it might be better to suggest throwing it in with yours.

I've added the `- 1`, but sometimes the button is still visible for some time due to this:

https://github.com/jarnedemeulemeester/findroid/blob/0a1582f47b33147c6c797c76f674bdedad4a4909/player/video/src/main/java/dev/jdtech/jellyfin/viewmodels/PlayerActivityViewModel.kt#L260

AbandonedCart:
Subtracting the 1 should force it to trigger the end a second early, which should offset performing the check each second well enough for most cases. There's still the possibility for a region of error less than a second long, but the only fix would be to lower that check and start eating away at the memory.

cd16b:
> sometimes the button is still visible

Fixed, if you want to test it let me know if it works.

AbandonedCart:
The only other thing that appears to be missing is that when the skip button is not set to run the full length of the intro, it is tied to the player controls. I think not reappearing may be why the issue was raised for the plugin to support full duration display.

That one is not a one-liner, though, so I will download the source and take a look before commenting further.

jarnedemeulemeester:
Okay so I took a quick look at this.

The next version of Findroid will target Jellyfin 10.9 so only jumoog's version will work. Knowing that I would completely migrate to the `IntroSkipperSegments` API call for both intros and credits. Doing that we would also be able to reuse more code for both intros and credits. 
Maybe with some generic `SkippableSegment` class with an enum property called `type` of type `SkippableSegmentType` which can be `Intro` or `Credit`. 
By doing that we can also store those in a single database table for offline access.

I have not taken a detailed look at the code or the API from jumoog's version but I think reworking the intro skip to something more generic like segment skipping would be ideal.

Also, I have not tested this yet, but I would like to include credit skipping in the next release. But no need to rush this since I have quite some more work to do before the next release :smile:

cd16b:
I changed from Intro and Credit to a generic FindroidSegment:
```
FindroidSegment(
            type = String,
            skip = Boolean,
            startTime = Double,
            endTime = Double,
            showAt = Double,
            hideAt =Double,
        )
```
The problem is with getting it from the Intro Skipper API, so currently the code isn't optimized.
If you have any ideas on how to improve it, please let me know.

jarnedemeulemeester:
I started looking at this, but have yet to review all of the changes.
I've changed it to load the segments at the beginning of each episode instead of all at once. 

Other than that I have a question about the `FindroidSegment`. Why does it have a property `skip`? It seems like it is always set to `true`?

cd16b:
The `skip` property can be removed, it is never used, I forgot to remove it.

jarnedemeulemeester:
Made some more changes:
- Restyled the buttons to the original outlined style with semi transparent background
- Changed the Continue watching button to a X icon button
- Also display the X button for intros
- Keep the buttons visible when the player controls are visible (I see no reason to hide them)

jarnedemeulemeester:
More changes:
- Introduce enum `FindroidSegmentType` to specify the type instead of using a string
- Use a map to retrieve the segments from the API and simplify the conversion to `FindroidSegment`s
- Save each segment in a row and it's properties in columns instead of using a single json

jarnedemeulemeester (approved):

jarnedemeulemeester:
Made some final changes. 
If everything looks good to you I think we can merge this 🙂

cd16b:
> Made some more changes:
> 
> * Keep the buttons visible when the player controls are visible (I see no reason to hide them)

For these, I copied the implementation of the main streaming apps which have different behaviour for intro and credits:
- **Intro** button is visible for n seconds, after that it is visible only if the player controls are visible
- **Credits** button is always visible unless the player controls are visible, so you don't have more buttons for the same function `Skip credits` and `Next`, but using this logic means overwriting IntroSkipper logic (maybe we can use the same logic as `Intro`)


Everything else looks fine 🙂👍

jarnedemeulemeester:
Let's follow your `Intro` logic. The button will be visible for n seconds, after that it's only visible when the player controls are shown. Let's do this for both the `Intro` and `Credits`.

I will be implementing that change now.

Btw I made an other pretty big change. I changed the `Skip credits` button to actually only skip the credits and not move on to the next episode. This is because sometimes there are extra scenes after the credits which a viewer would want to watch. I got this idea from Crunchyroll (I don't have access to other streaming services), and it makes sense in my opinion. `Skip intro` skips the intro and `Skip credits` skips the credits. 
It also makes the code a little less complex :slightly_smiling_face:

PR #640, opened 2024-01-28
fix: pip zoom start player maximized

cd16b:
Sorry, I found one more problem.

### DIFF
--- a/app/phone/src/main/java/dev/jdtech/jellyfin/PlayerActivity.kt
+++ b/app/phone/src/main/java/dev/jdtech/jellyfin/PlayerActivity.kt
@@ -55,6 +55,7 @@ class PlayerActivity : BasePlayerActivity() {
     private var playerGestureHelper: PlayerGestureHelper? = null
     override val viewModel: PlayerActivityViewModel by viewModels()
     private var previewScrubListener: PreviewScrubListener? = null
+    private var wasZoom: Boolean = false
 
     private val isPipSupported by lazy {
         // Check if device has PiP feature
@@ -308,6 +309,7 @@ class PlayerActivity : BasePlayerActivity() {
         binding.playerView.useController = false
         binding.playerView.findViewById<Button>(R.id.btn_skip_intro).isVisible = false
 
+        wasZoom = playerGestureHelper!!.isZoomEnabled
         playerGestureHelper?.updateZoomMode(false)
 
         // Brightness mode Auto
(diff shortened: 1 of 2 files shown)


Natanel-Shitrit:
Why aren't you just tracking `isZoomEnabled` inside `updateZoomMode` instead of updating it in each place you call `updateZoomMode`?

Natanel-Shitrit:
`isZoomEnabled` is not used anywhere, it can be discarded entirely.

cd16b:
> Why aren't you just tracking `isZoomEnabled` inside `updateZoomMode` instead of updating it in each place you call `updateZoomMode`?

You are right. To avoid future issues, I will update `isZoomEnabled` inside `updateZoomMode`.

Natanel-Shitrit:
> `isZoomEnabled` is not used anywhere, it can be discarded entirely.

Please note that it can be just discarded because this value is not being used anywhere.

cd16b:
> `isZoomEnabled` is not used anywhere, it can be discarded entirely.

It's used in line 332: `isZoomEnabled = scaleFactor > 1` to call `updateZoomMode`.
I removed the 'private' modifier so I could use it when exiting pip mode in the PlayerActivity.

Natanel-Shitrit:
> > `isZoomEnabled` is not used anywhere, it can be discarded entirely.
> 
> It's used in line 332: `isZoomEnabled = scaleFactor > 1` to call `updateZoomMode`. I removed the 'private' modifier so I could use it when exiting pip mode in the PlayerActivity.

it's not used, it's being assigned a value, you can just pass `scaleFactor > 1` to the function or create a local `val` if you want a variable for readability.

cd16b:
> > > `isZoomEnabled` is not used anywhere, it can be discarded entirely.
> > 
> > 
> > It's used in line 332: `isZoomEnabled = scaleFactor > 1` to call `updateZoomMode`. I removed the 'private' modifier so I could use it when exiting pip mode in the PlayerActivity.
> 
> it's not used, it's being assigned a value, you can just pass `scaleFactor > 1` to the function or create a local `val` if you want a variable for readability.

Ok, but before entering picture-in-picture mode, I need to save the zoom state. This is because when using the MPV player, I cannot use playerView.getGlobalVisibleRect() and it behaves strangely when the video is zoomed.

cd16b:
> > Why aren't you just tracking `isZoomEnabled` inside `updateZoomMode` instead of updating it in each place you call `updateZoomMode`?
> 
> You are right. To avoid future issues, I will update `isZoomEnabled` inside `updateZoomMode`.

I'm reverting the latest commit cause I have to save the zoom state before changing it.

cd16b:
Now tracking `isZoomEnabled` inside `updateZoomMode`

PR #644, opened 2024-01-30
fix: media folders are invisible on 10.9.0

iFraan:
Seems to be the issue originated with the migration of CollectionType to enums from upstream ([#9764](https://github.com/jellyfin/jellyfin/pull/9764) & [#10702](https://github.com/jellyfin/jellyfin/pull/10702))

Lowercasing it was enough. 

Draft since 10.9.0 isn't stable yet, but i don't think merging this would cause any issue.

### DIFF
--- a/data/src/main/java/dev/jdtech/jellyfin/models/CollectionType.kt
+++ b/data/src/main/java/dev/jdtech/jellyfin/models/CollectionType.kt
@@ -27,7 +27,7 @@ enum class CollectionType(val type: String) {
             }
 
             return try {
-                entries.first { it.type == string }
+                entries.first { it.type == string.lowercase() }
             } catch (e: NoSuchElementException) {
                 defaultValue
             }


jarnedemeulemeester:
I don't think this is necessary. In jellyfin/jellyfin#10702 the enums got converted to lowercase again.
Are you testing the latest unstable version?

iFraan:
Thank you for replying!

I'm testing one of the the latest unstable versions (jan-26-2024), and the Jellyfin API returns PascalCase strings for these enums.

Also i think

> I tried multiple methods of using a lowercasing string enum converter but wasn't able to find a method that worked fully.

means that they won't change it and maybe it's a gonna become a 'breaking change' when updating to 10.9.0 (i could be wrong tho)

Probably next 10.8.x release won't include this anyways, and 10,9.0 isn't close enough (hence the draft nature of this PR), but 
worth noting that it exists.

iFraan:
Just updated to the latest unstable server.
Default app works as intended, closing this PR as is no longer necessary.

PR #648, opened 2024-02-02
feat: automatic PiP animation

Natanel-Shitrit:
Make transitions to PiP mode smoother from gesture navigation: 

https://github.com/jarnedemeulemeester/findroid/assets/65548905/152a7a31-9d71-4bd9-97cf-a5e9b8ca9f91

> Demo from: https://developer.android.com/develop/ui/views/picture-in-picture

For Android 12 and newer, only tested on android 14.

### DIFF
--- a/app/phone/src/main/java/dev/jdtech/jellyfin/PlayerActivity.kt
+++ b/app/phone/src/main/java/dev/jdtech/jellyfin/PlayerActivity.kt
@@ -277,7 +277,7 @@ class PlayerActivity : BasePlayerActivity() {
         }
     }
 
-    private fun pipParams(enableAutoEnter: Boolean = false): PictureInPictureParams {
+    private fun pipParams(enableAutoEnter: Boolean = viewModel.player.isPlaying): PictureInPictureParams {
         val displayAspectRatio = Rational(binding.playerView.width, binding.playerView.height)
 
         val aspectRatio = binding.playerView.player?.videoSize?.let {
--- a/app/phone/src/main/java/dev/jdtech/jellyfin/PlayerActivity.kt
+++ b/app/phone/src/main/java/dev/jdtech/jellyfin/PlayerActivity.kt
@@ -160,6 +160,10 @@ class PlayerActivity : BasePlayerActivity() {
                                 speedButton.imageAlpha = 255
                                 pipButton.isEnabled = true
                                 pipButton.imageAlpha = 255
+
+                                if (appPreferences.playerPipGesture) {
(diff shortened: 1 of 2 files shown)


cd16b:
Hi, I added PiP support last year.
However, I removed the `setAutoEnterEnabled()` feature because of two problems.
- It should only be enabled when the video is playing, like on YouTube. So, you have to update it every time the playback status changes, and I wasn't able to make it work correctly. 
- MPV player: for a smoother transition, it's better to use the `getGlobalVisibleRect()` function, but unfortunately, it's not available for MPV player at the moment (that's why it's using that  long unnecessary code to get `SourceRectHint`), but it can be used for Exoplayer.

Last I recommend calling 
```
binding.playerView.useController = false
binding.playerView.findViewById<Button>(R.id.btn_skip_intro).isVisible = false
playerGestureHelper?.updateZoomMode(false)
```
inside `pictureInPicture()` to hide all player controls before the animation begins, at least when using the PiP button.

Natanel-Shitrit:
> - It should only be enabled when the video is playing, like on YouTube. So, you have to update it every time the playback status changes, and I wasn't able to make it work correctly. 

It's working fine when paused, am I missing something? 😅

> - MPV player: for a smoother transition, it's better to use the `getGlobalVisibleRect()` function, but unfortunately, it's not available for MPV player at the moment (that's why it's using that  long unnecessary code to get `SourceRectHint`), but it can be used for Exoplayer.

Is that something that effects the animation?
MPV player is animated correctly with the current state of the code.

> Last I recommend calling 
> ```
> binding.playerView.useController = false
> binding.playerView.findViewById<Button>(R.id.btn_skip_intro).isVisible = false
> playerGestureHelper?.updateZoomMode(false)
> ```
> inside `pictureInPicture()` to hide all player controls before the animation begins or at least when using the PiP button.

I see what you are talking about, the callback is firing only after the state change, but it's really hard to notice, so I guess it's not really a big deal.

jarnedemeulemeester:
Okay so I rebased this on top of `main` and fixed the merge conflict with the recent zoom fix.

I also added the fix for disabling the pip gesture when content is paused. I took the implementation from @cd16b but modified it a little bit.

While using a `preparePip` function before calling `enterPictureInPictureMode` creates a smoother animation by first disabling the controls and resetting the zoom mode. I don't think it's worth it because in my opinion it make the code less readable.

Also fixed a crash when gestures are disabled and entering pip using the button.

This PR + my changes have been tested on Android 9 and Android 14.

Issue #469, opened 2023-08-14
Linebreak (<br>) does not render in findroid


compile10:
**Describe the bug**
Line break html tags \<br> do not render in show summaries or episode summaries. Summary just shows linebreak tag \<br> instead of actually adding a new line.
**To Reproduce**
1. Add \<br> to text in a show or episode summary
2. Open show summary in app

**Expected behavior**
Line break to be rendered like in jellyfin web client.

**Device info (please complete the following information):**
 - Device: Google Pixel Fold
 - Android version: 13
 - App version: 0.12.0 (20)

GreatestFool:
I presume this is because it doesn't parse any HTML, but I'd like to see this implemented too. Even if full support for HTML is out of scope, I think it would be nice if it could at least match certain tags like bold, italics and line breaks. Hopefully also anchors, because I have a few that link to their original source for YouTube videos and merged episodes (e.g. E01-E03).

joshuaavalon:
I have looked up the document. It should be a rather simple fix.

https://github.com/jarnedemeulemeester/findroid/blob/07c58accf00e0d5dd497f1e74a0138639777fcd0/app/phone/src/main/java/dev/jdtech/jellyfin/adapters/EpisodeListAdapter.kt#L48

```kt
binding.episodeOverview.text = if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
    Html.fromHtml(html, Html.FROM_HTML_MODE_COMPACT)
} else {
    Html.fromHtml(html)
}
```

android.text.Html is included in the standard library. It just needed to wrap overview's  `TextView.setText` with  `Html.fromHtml`.

@jarnedemeulemeester Are there any concerns that this cannot be implemented? Or should I create a PR for you?

PR #713, opened 2024-04-02
feat: support html in media description

leekleak:
Before: 

After:

fixes: #469


jarnedemeulemeester:
My only concern is the performance of `fromHtml`. 
However media item descriptions are often not very long so the performance impact should be rather small.
I did some quick tests with and without `fromHtml` and it seems fine.

Thanks!

PR #714, opened 2024-04-03
Settings navigation fixes

leekleak:
The pr addresses all settings sub menus except "Servers", "Users" and "Addresses" (Because they are already working as expected).
Changes for all the other sub menus:
- Entering and exit animations are now consistent across all side menus.
- Pressing the back button in the header returns to the main settings menus instead of returning to home.
- Header label now correctly states the sub menu name instead of just saying "Settings"

P.S. Previously the animations for the changed sub menus were "slide in/out" whereas now they're "fade in/out". Not sure if this is better, maybe I should change them all to be "slide in/out"?


Panda260:
Oh very nice especially when you go back with the arrow you don't end up on the home screen but actually go back.

PR #720, opened 2024-04-08
feat: add support for "homevideos" collections

hax0r31337:
`homevideos` collections is for libraries created with `Photos` type which is for unorganized folders

image


chiragkrishna:
came to check why Home Videos were not showing up and came across this. please merge and make "Home Videos and Photos" available as its a simple merge.

PR #723, opened 2024-04-12
Automatically connect to reachable address on app launch

leekleak:
I'm gonna be honest. This may just be the most disgusting piece of code I've written. There's:
- Code duplication.
- Unforeseen consequences (The app will not leave the loading screen if all addresses are unreachable and I'm sure there are many more).
- General ugliness.

What (I think) the pr does is it pings all registered server addresses and connects to the one with the fastest response, so it saves a couple of clicks if you often use something like Tailscale to access the app from outside the local network.

If I know it's bad, why am I still creating this PR?
I... I didn't think about that actually.
Honestly I just hope to get some guidance on how to do this better as I am a total android development newbie. I hope I don't sound like a troll though :P

### DIFF
--- a/app/phone/src/main/java/dev/jdtech/jellyfin/MainActivity.kt
+++ b/app/phone/src/main/java/dev/jdtech/jellyfin/MainActivity.kt
@@ -2,6 +2,7 @@ package dev.jdtech.jellyfin
 
 import android.os.Bundle
 import android.view.View
+import android.widget.Toast
 import androidx.activity.viewModels
 import androidx.appcompat.app.AppCompatActivity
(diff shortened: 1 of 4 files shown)


Panda260:
I think the idea is nice but I think it makes absolutely no sense, especially if you have a second test jellyfin server at home and you want to connect to a server that is not on your local network.

leekleak:
@Panda260 How so? Alternate addresses are, well alternate addresses for the same server. This pr doesn't force you to switch servers, it just checks if the server that is opened on startup is reachable and if it is, through which address.

PR #738, opened 2024-04-24
Show thumbnails even when server is unreachable

### LINKED ISSUE #211: [Improvements]  The downloads do not show covers
The downloads do not show the cover of Seasons, Movies or chapters supposedly having the photos cached in the app.
photo_2022-12-04_02-24-42

### LINKED ISSUE #525: No thumbnails in Offline Mode
**Describe the bug**
No thumbnails of downloaded content.

**To Reproduce**
1. Download any media while being connected.
2. Disconnect from the server.
3. Open Findroid Application.

**Expected behavior**
Thumbnails of downloaded videos should also be available. Text and rating is present but not images.

**Screenshots**
Screenshot_20230926-184530
Screenshot_20230926-184538

 - Device: Motorola G32
 - Android version: 12
 - App version: 0.13.0(21)
 - Jellyfin version: 10.8.8

### LINKED ISSUE #800: Not Showing Metadata (Images) While Offline for Downloaded Files

leekleak:
This PR fixes an issue where if the server is unreachable (being in offline mode does not suffice, you have to disable the wifi/data manually) only grey placeholders are rendered.

Fixes #525
Fixes #211
Fixes #800

### DIFF
--- a/app/phone/src/main/java/dev/jdtech/jellyfin/BaseApplication.kt
+++ b/app/phone/src/main/java/dev/jdtech/jellyfin/BaseApplication.kt
@@ -69,7 +69,7 @@ class BaseApplication : Application(), Configuration.Provider, ImageLoaderFactor
                         chain.proceed(
                             chain.request().newBuilder()
                                 .header("Cache-Control", "stale-if-error")
-                                .build()
+                                .build(),
                         )
                     }
                     .build()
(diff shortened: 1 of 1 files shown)


jarnedemeulemeester:
I don't think this is the way to go. This change makes it so that images are cached forever. Changing the image on the server would never change it on the client.

A better way to fix this would be to store the images locally for downloaded media items.

PR #741, opened 2024-04-29
Fix wrong "Date Played" sort for TV shows

Hjaelp:
For series content, Jellyfin uses "[SeriesDatePlayed](https://github.com/jellyfin/jellyfin/blob/2cbef3aa383b05e51b76fb399a2a2fded56518e5/Jellyfin.Data/Enums/ItemSortBy.cs#L146)" to sort by last played rather than the current "DatePlayed".

Tested it, and I'm getting the same sort result in findroid and jellyfin with this change, but I'm not sure the best way to do this change. I was going to create a separate enum, but this is the only sort property that has has this issue.

### DIFF
--- a/core/src/main/java/dev/jdtech/jellyfin/dialogs/SortDialogFragment.kt
+++ b/core/src/main/java/dev/jdtech/jellyfin/dialogs/SortDialogFragment.kt
@@ -51,7 +51,12 @@ class SortDialogFragment(
                             sortByOptions,
                             currentSortBy.ordinal,
                         ) { dialog, which ->
-                            val sortBy = sortByValues[which]
+                            val sortBy =
+                                if (libraryType == CollectionType.TvShows && sortByValues[which] == SortBy.DATE_PLAYED) {
+                                    SortBy.SERIES_DATE_PLAYED
+                                } else {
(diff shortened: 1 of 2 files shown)


jarnedemeulemeester:
Damn, that's really dumb. Why isn't `DatePlayed` just the date played of a series instead of creating a separate value `SeriesDatePlayed`... Looks like something that was inherited from Emby back in the day. Or maybe I'm missing the reason.

IMO the ideal solution would be for Jellyfin to make `DatePlayed` the actual date played of a series.

But I suppose we can also implement a fix in the mean time. 

Your solutions does not completely work. When Date Played is selected for a series library it is not visibly selected in the sort dialog. And also when selecting it for a series library and navigating to a movies library it does not sort the movies properly.
I have an other fix inspired by yours which I will push to this branch shortly.

jarnedemeulemeester:
@Hjaelp please review and test my changes if they still work for you :slightly_smiling_face:

Hjaelp:
Thank you, I completely forgot that the sort is saved. I'll test it on my phone and tablet today and Wednesday.

Hjaelp:
Yep, tested it for a bit and everything worked perfect.

PR #767, opened 2024-05-30
fix(nav): avoid navigation crashes

### LINKED ISSUE #758: Crash when double tapping elements that initiate navigation
**Describe the bug**
occasionally findriod crashes while leaving the app

**To Reproduce**
play a video then exit to the main menu

**Screenshots**

https://github.com/jarnedemeulemeester/findroid/assets/93036561/833f3f93-db83-4393-a844-4a915dfc6510

 - Device: [A53]
 - Android version: [14]
 - App version: [0.14.2]
 - Jellyfin version: [10.9.2]

**Additional context**
I'm using mpv player

Natanel-Shitrit:
Introduce a `safeNavigate` function for navigation controller, to prevent crashes when a navigation fails.

## Related Issues:
- Fixes #758

### DIFF
--- a/app/phone/src/main/java/dev/jdtech/jellyfin/fragments/AddServerFragment.kt
+++ b/app/phone/src/main/java/dev/jdtech/jellyfin/fragments/AddServerFragment.kt
@@ -17,6 +17,7 @@ import androidx.navigation.fragment.findNavController
 import dagger.hilt.android.AndroidEntryPoint
 import dev.jdtech.jellyfin.adapters.DiscoveredServerListAdapter
 import dev.jdtech.jellyfin.databinding.FragmentAddServerBinding
+import dev.jdtech.jellyfin.utils.safeNavigate
 import dev.jdtech.jellyfin.viewmodels.AddServerEvent
 import dev.jdtech.jellyfin.viewmodels.AddServerViewModel
 import kotlinx.coroutines.launch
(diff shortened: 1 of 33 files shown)


Natanel-Shitrit:
Basically this:
https://github.com/jarnedemeulemeester/findroid/blob/e3e1a60001d2e7137a2b9b99c4575b9ccea989a4/app/phone/src/main/java/dev/jdtech/jellyfin/utils/Extensions.kt#L20-L26
And replace all `navigate` calls to `safeNavigate`.

Natanel-Shitrit:
I've reverted the changes for the TV version since I can't really test if the original problem also consists in the TV version.

jarnedemeulemeester:
This looks like a good improvement.

However, I've been slowely rewriting the UI in Jetpack Compose and I'm not sure the same problem appears in that framework. 
The plan is to have all UI (except the player) be rewritten in Compose by the next release.
If that were to fail, I will merge this for the next release.

One more question, are we sure this fixed the issue where the app crashes when exiting the player?

Natanel-Shitrit:
I'm all for Jetpack compose 😁

About the player - according to the the person who opened the issue, yes.

It also looks like the crash didn't really happen because he closed the player but managed to click on an episode while it was loading another episode already.

jarnedemeulemeester:
I will merge this since the compose UI refresh is still for from finished

Issue #43, opened 2021-10-10
Allow transcoding of multimedia content.


JulioNahum:
**The ability to choose the quality of playback when choosing a movie or series is urgently needed.**

**I think this feature would be the most essential since it would be the first jellyfin client for android that allowed content transcoding.** _I AM GRATEFUL FOR YOUR WORK and I WOULD LIKE YOU TO INCLUDE IT AS SOON AS POSSIBLE._ 

**I know we all have our lives and work, and you do too, but you have shown that you have done an excellent job creating this client. And I hope this client is the official Jellyfin for mobile needs. With this feature, the application would level up and be above EMBY.**

SmearyThrower:
any chance this ever coming to findriod🤞🏻

jramseygreen:
I, and many other jellyfin users I know, love this project but find the lack of transcoding a complete deal breaker. I really hope it's supported soon and this will become our jellyfin daily drive client!

isZYKerman:
I really need the transcode feature. When I commute, my network becomes so poor that it cannot support high bitrate video. ☹️ I really want to see this feature in the app as soon as possible.

mattalxndr:
Yeah hopefully this feature makes the list at some point. It's a great player though, when I'm unmetered.

BillCipher-exe:
Yes please add this feature

Bhuemann:
Yes I don't need any more features but this one

KdntNinja:
This would also be useful as my home network doesn't have a high upload speed so streaming outside doesn't work without transcoding.

l0stpt:
Is this going to happen in the future? 

Would be really cool since the official android jellyfin client can't play hevc files, only the web player but the web player sucks.

Also thanks for this amazing software I love it and use it daily.

PR #791, opened 2024-07-18
feat: Transcoding stream in player selection + Transcoded downloads

### LINKED ISSUE #775: Transcode downloads to save storage space
My tablet does not need to save a 4k version of the movie.

Saving a transcoded version with lower resolution (picked at time of download with a default setting) would be nice.

nomadics9:
- Transcoded stream support (Auto(adaptive) - Original - 1080p - 720p - 480p - 360p)
- Transcoded download support (Original - 1080p - 720p - 480p - 360p)

- [x] Fully working transcoded stream
- [x] Embedded subtitle in transcoded stream
- [x] Transcoded download (Original - 720p - 480p - 360p)
- [x] Adding embedded subtitle tracks externally when playing offline 
- [x] App preference for download quality
- [x] Dialog to pick download quality when download if default is not set 

## TODO:

- [x] Enum  -> [commit](https://github.com/jarnedemeulemeester/findroid/pull/791/commits/6dded2e72641868731d4716e6b35e906496b9799) [commit](https://github.com/jarnedemeulemeester/findroid/pull/791/commits/7adcc50d750d82e08104987e97665ab9c65c85f9)
- [x] Strings + icon name -> [commit](https://github.com/jarnedemeulemeester/findroid/pull/791/commits/c79342523be289a34d0bfd1a143e3c3abf3ecd88)
- [x] Rework getting OriginalHeight for quality selection dialog -> [commit](https://github.com/jarnedemeulemeester/findroid/pull/791/commits/21ae815223643a8dae0fa28cdf2e4ca00486b07a)
- [x] Add option to choose between h265(HEVC) or h264 to transcode into -> [commit](https://github.com/jarnedemeulemeester/findroid/pull/791/commits/8482df97334301057081a735491b00bd8d84e1c5)
- [x] Code clean up -> [commit](https://github.com/jarnedemeulemeester/findroid/pull/791/commits/5609f7368d34ed6a7067c29d35dfdcc3eca03f00)

#### Enum commits changes [1](https://github.com/jarnedemeulemeester/findroid/pull/791/commits/6dded2e72641868731d4716e6b35e906496b9799)     [2](https://github.com/jarnedemeulemeester/findroid/pull/791/commits/7adcc50d750d82e08104987e97665ab9c65c85f9)

- New Enum "VideoQuality"
- Add Resolution 1080p
- Bitrates values change -> 1080p - 8Mbps | 720p - 3Mbps | 480p - 1.5Mbps | 360p 800Kbps 
- All transcodes to H264 and AAC-128k
- Bugfix causing unprivileged users not able to change stream quality
- Show selected quality selection dialog

## Apks 

- [x] [Here](https://github.com/nomadics9/ananas/actions/runs/10023633669)

#### Closes
#775 
#43


core/src/main/java/dev/jdtech/jellyfin/utils/DownloaderImpl.kt
```diff
@@ -82,20 +85,40 @@ class DownloaderImpl(
...
-                    val request = DownloadManager.Request(source.path.toUri())
-                        .setTitle(item.name)
-                        .setAllowedOverMetered(appPreferences.downloadOverMobileData)
-                        .setAllowedOverRoaming(appPreferences.downloadWhenRoaming)
-                        .setNotificationVisibility(DownloadManager.Request.VISIBILITY_VISIBLE_NOTIFY_COMPLETED)
-                        .setDestinationUri(path)
-                    val downloadId = downloadManager.enqueue(request)
-                    database.setSourceDownloadId(source.id, downloadId)
-                    return Pair(downloadId, null)
+                    if (appPreferences.downloadQuality != "Original") {
```
Jcuhfehl, line 88: This string is a sort of [magic number](https://en.wikipedia.org/wiki/Magic_number_(programming)) (i'm not sure if there's a better term which also encompasses strings). You should probably replace this with an enum

app/phone/src/main/java/dev/jdtech/jellyfin/fragments/MovieFragment.kt
```diff
@@ -192,72 +192,82 @@ class MovieFragment : Fragment() {
...
-            }
+                handleDownload()
         }
 
         binding.peopleRecyclerView.adapter = PersonListAdapter { person ->
             navigateToPersonDetail(person.id)
         }
     }
 
+    private fun handleDownload() {
```
Jcuhfehl, line 203: A lot of this code is the same between the `EpisodeBottomSheetFragment` and `MovieFragment`. Since a lot of it is UI logic I'm not sure it could be avoided, but I think it'd be worth checking if there's a way to reduce the duplication.

core/src/main/java/dev/jdtech/jellyfin/utils/DownloaderImpl.kt
```diff
@@ -263,4 +390,66 @@ class DownloaderImpl(
...
         }
     }
+
+    private suspend fun getTranscodedUrl(
+        itemId: UUID,
+        quality: String,
+    ): Uri? {
+        val maxBitrate =
+            when (quality) {
+                "720p" -> 2000000 // 2 Mbps
```
Jcuhfehl, line 400: Same "magic number" problem, I'd prefer an enum here

core/src/main/res/drawable/ic_quality.xml
Jcuhfehl, line 1: The file names for icons should be the same as their names on [lucide.dev](https://lucide.dev/)

app/phone/src/main/java/dev/jdtech/jellyfin/PlayerActivity.kt
```diff
@@ -342,6 +347,23 @@ class PlayerActivity : BasePlayerActivity() {
         } catch (_: IllegalArgumentException) { }
     }
 
+    private fun showQualitySelectionDialog() {
+       val height = viewModel.getOriginalHeight() // TODO: rewrite getting height stuff I don't like that its only update after changing quality
+        val qualities = when (height) {
+            0 -> arrayOf("Auto", "Original - Max", "720p - 2Mbps", "480p - 1Mbps", "360p - 800kbps")
```
Jcuhfehl, line 353: This cannot be localized because the strings are hardcoded. You should use `strings.xml` for this so the strings can be translated in weblate

core/src/main/res/values/string_arrays.xml
```diff
@@ -25,4 +25,16 @@
         <item>audiotrack</item>
         <item>opensles</item>
     </string-array>
+    <string-array name="quality_entries">
```
Jcuhfehl, line 28: Same thing here: these item names cannot be localized, so they should refer to `strings.xml`

core/src/main/res/xml/fragment_settings_downloads.xml
```diff
@@ -9,4 +9,16 @@
         android:defaultValue="false"
         app:key="pref_downloads_roaming"
         app:title="@string/download_roaming" />
+<!-- TODO: Strings -->
+    <ListPreference
+        android:key="pref_downloads_quality"
+        android:title="Download Quality"
+        android:defaultValue="Original"
```
Jcuhfehl, line 16: Same localization issue here

Jcuhfehl (changes requested):
Looks nice (although I didn't test it yet). I noticed a few things that could be improved.
Also I think adding a transcoded 1080p (maybe even a 4k one) would be useful. People may store 4k BDRemuxes, so getting a stream that's not full blu-ray bitrate would often be needed in that case.

UriSheinman:
Graphic Subtitles don't work when transcoding, that is quite an issue...

nomadics9:
> Graphic Subtitles don't work when transcoding playback, that is quite an issue... Edit: In general subtitles and track selection is kinda all over the place, sometimes it shows options sometimes it doesn't, sometimes it shows track's language and sometime it doesn't. Also there's no indicator in the quality selection as to what is selected. There should be a checkmark or something...

hey, thanks for testing! for PSGSUB I noticed I never get a deliveryurl for this specific codec only. not sure why the path is always null. However there is one more thing i would like to try. 

As for audio tracks there is no logic to handle audio tracks yet. I havent even thought about it .I will need to have a look.

app/phone/src/main/java/dev/jdtech/jellyfin/fragments/MovieFragment.kt
nomadics9: Those are 2 different UI fragments might be by design to separate the two when adjustment is needed so I left it as is. I will mark it as resolved.

MarkusTieger:
If it gets transcoded anyway, why not just add a way to deselect certain audio tracks from being downloaded?
Also would be nice to have a way to change that: "All transcodes to H264 and AAC-128k"
(haven't tried the apk, just assumptions from the issue description)

jarnedemeulemeester:
Just to let you know, I have been looking at this PR from time to time.
Once I have a little more time to spare I will take a look at it in more detail.

NickLewanowicz:
Any chance this gets merged soon? Streamyfin has a similar transcode download feature but IMO Findroid is a much better UI. 

It was posted on [this jellyfin feature request](https://features.jellyfin.org/posts/57/support-download-of-transcoded-files) that has been open for 5 years with over 300 upvotes. I imagine many of those people would love to see they have multiple mobile app options for transcoding offline downloads.

viggy96:
Does this PR tell the server to do tone mapping for HDR content or does it tell the server to passthrough the HDR metadata for the client to deal with? I would prefer the latter.

nomadics9:
> Just to let you know, I have been looking at this PR from time to time.
> Once I have a little more time to spare I will take a look at it in more detail.

Sorry I haven't gotten back to you sooner. It's been a while since I have done this. A small caveat I never wrote Kotlin before or after this PR. The feature might need a bit of tidying. It works. Since last time I used it. I haven't been updating my fork since then.

But since it's a big PR I do understand that it's a pain to work with. If you need any help please let me know! I will need to revisit it though.

nomadics9:
> Does this PR tell the server to do tone mapping for HDR content or does it tell the server to passthrough the HDR metadata for the client to deal with? I would prefer the latter. 

If I remember correctly. It uses your server settings. I was only telling the server to transcode into what, h265 or h264 which you can pick in settings. Video bitrate and audio bitrate which you can pick in video player. Nothing else regarding HDR.

ishanjain28:
Hii, here to voice my support for this pr.
I have friends who use findroid and transcoded streams will be a huge help!

Speiger:
Any update on this?
Because it looks pretty dead...

PR #870, opened 2024-10-27
TV playback controls: add fast seeking and enable track selection without pausing

voodoos:
Hi @jarnedemeulemeester thanks a lot for your work ! I am using the Android TV version of Findroid because the playback is much more robust than the official one. But I found the controls to be a bit finicky... here is a proposition to improve them a bit, adding the possibility of fast seeking and track switching without pause. I hope this is of interest to you ! It is my first time using Compose so I would be happy to receive feedback and improve my code.

This PR proposes an addition to the current playback controls to allow seeking and changing tracks without pausing the playback. 

The current controls are especially tedious to use for seeking:

1. First press the <kbd>Enter</kbd> key to pause to show the controls. This also pauses the playback.
2. Navigate to the seeker (right then bottom).
3. Activate seeking by pressing <kbd>Enter</kbd>.
4. Seek by pressing multiple times <kbd>Right</kbd>  or <kbd>Left</kbd>.
5. Validate by pressing <kbd>Enter</kbd>.
7. Press <kbd>Enter</kbd> to resume the playback.

This PR adds two new behaviors:

1. Fast-seeking by pressing <kbd>Right</kbd> or <kbd>Left</kbd>  keys during playback while the controls are not shown.
2. During, playback, pressing the <kbd>Up</kbd>  shows the controls _without pausing the playback_ which allows changing the audio or subtitle tracks without pausing. 

However, I had to inhibit several current behaviors to achieve that:
- When seeking, the player state briefly switches to `paused`  while buffering. But this is not a pause to which the UI should react. In commit dc1aac3d6733b01b9407cc63a470fedd43b2a3dc I make UI pause-state independent from the player's own state so that we can have finer behaviors.
- The `Box` containing the video surface has custom bindings for the D-pad keys: on <kbd>Enter</kbd> it pauses playback and shows the controls. These actions were still performed while the controls were shown (without a negative impact). In commit a8461c1914b383669f9124cf5dab2a99a42cbfc5 I restrict these actions to trigger only if the player box is in focus. This prevent fast-seek on <kbd>Right</kbd> and <kbd>Left</kbd> to trigger while navigating in the controls.
-  Opening a track selection dialog pauses the player's activity which used to pause the playback. In c30bc52e447b1c229c7f92c7a8422f109d9dd898 I disable that behavior to allow track switching without pausing.

### DIFF
--- a/app/tv/src/main/java/dev/jdtech/jellyfin/ui/PlayerScreen.kt
+++ b/app/tv/src/main/java/dev/jdtech/jellyfin/ui/PlayerScreen.kt
@@ -260,8 +260,6 @@ fun VideoPlayerControls(
             ) {
                 VideoPlayerMediaButton(
                     icon = painterResource(id = R.drawable.ic_speaker),
-                    state = state,
-                    isPlaying = isPlaying,
                     onClick = {
                         val tracks = getTracks(player, C.TRACK_TYPE_AUDIO)
                         navigator.navigate(VideoPlayerTrackSelectorDialogDestination(C.TRACK_TYPE_AUDIO, tracks))
@@ -269,8 +267,6 @@ fun VideoPlayerControls(
                 )
                 VideoPlayerMediaButton(
                     icon = painterResource(id = R.drawable.ic_closed_caption),
-                    state = state,
-                    isPlaying = isPlaying,
                     onClick = {
                         val tracks = getTracks(player, C.TRACK_TYPE_TEXT)
                         navigator.navigate(VideoPlayerTrackSelectorDialogDestination(C.TRACK_TYPE_TEXT, tracks))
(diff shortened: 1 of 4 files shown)


voodoos:
I see that linting fails to to some style issues (and missing imports). Is there anyway for me to configure my Android Studio to enforce these styles ?

jarnedemeulemeester:
> I see that linting fails due to some style issues (and missing imports). Is there anyway for me to configure my Android Studio to enforce these styles ?

I don't know of a way to show the linting errors in Android Studio. However you can run `./gradlew ktlintFormat` to check and lint your code

jarnedemeulemeester:
I like the changes to open up the controls without pausing the playback.
But I am not sure if coupling `isPlaying` to the user interaction is a good idea. What if the playback get's pause by the mediasession from for example a phone or something?

voodoos:
> But I am not sure if coupling `isPlaying` to the user interaction is a good idea. What if the playback get's pause by the mediasession from for example a phone or something?

I think you're right, it's not completely satisfying right now, and I noticed some strange behavior sometimes when sending some inputs right after starting the videos that might be linked to that. I will try to rework it in the following weeks. I do think we need to have some kind of fast seeking that doesn't have to show the controls however, not sure how that can be achieved...

PR #871, opened 2024-10-27
Feat: 10.10 MediaSegments

### LINKED ISSUE #833: Skip intro button is tied to the media controls
**Describe the bug**
The skip intro button appears and disappears as part if the media controls are showing or not, though they should be separate.
**To Reproduce**
1. Watch a content with skip intro
2. Tap screen to show media control
3. Tap again to make the media control go away
4. The skip intro button disappears with the media controls

**Expected behavior**
The skip intro button should behave independently and not part of the media controls

**Screenshots**

https://github.com/user-attachments/assets/f8d2ddc3-51eb-4c02-ba03-85b1abc18995

 - Device: nothing phone 1
 - Android version: 14
 - App version: 0.15.2
 - Jellyfin version: 10.9.9

### LINKED ISSUE #834: Add the ability to choose if you want the skip credits to go to next episode or just skip credits
Non of my shows has a post credits scene so if it would be possible to add the option to just skip to next episode?

### LINKED ISSUE #851: Auto skip intro/credits toggle
It would be nice to have an option in the settings to automatically skip intros and credits instead of having to manually press the skip intro/credits button. As I understand it the intro/credit skip in Findroid simply uses the timestamp data from jumoog's plugin rather than using the plugin to handle the actual skipping.

**Describe alternatives you've considered**
I believe the default auto skip in the plugin settings for jumoog's intro skipper currently only works with the web player and players based off of that (e.g. official android app web player)

cd16b:
**Add** Jellyfin 10.10 MediaSegments

**Add** Auto Skip options: 
- never
- only in pip mode
- always

**Add** Skip to Next Episode if there is no post credits scene

**Fix**: Skip Button visible after being pressed
____________________________
### To do

- [x] Update preferences
- [x] Set default values
____________________________
closes #851 
closes #834
fix #833 
____________________________
Edit: ~I wrote this before Jellyfin 10.10.0, so it may need a few changes.~


jarnedemeulemeester:
Awesome work!
I will review this soon and try to include it in the next release

cd16b:
```kotlin
data class MediaItemSegments(
    val itemId: UUID,
    val segments: List<FindroidSegment>,
)
``` 

`itemId`  is used to avoid using previous item segments, especially when using auto skip.
Or we can add the `itemId` to the `FindroidSegment` itself, which is provided by the mediaSegmentsApi

PR #914, opened 2024-12-23
feat(film): migrate to compose: home

jarnedemeulemeester:
This migrates the home screen to compose.

The original xml layout files have been removed.

This also introduces the concept of modes. Right now this is just a folder in the root directory named `modes` containing a single `film` module. The goal is to in the future move all film (movies and tv shows) related code to that module.
Later we could then add other modes like `music`, `books` and `photos`.

## Screenshots
Screenshot_20241223-115152


Matix-Media:
I very much appreciate the larger search bar

Matix-Media:
I am wondering how the bottom navigation bar will look like with the settings button moved out of the quick actions

jarnedemeulemeester:
> I am wondering how the bottom navigation bar will look like with the settings button moved out of the quick actions

The settings button is now the user button in the search bar. It will still take you to the settings. Only the icon is changed.
Now showing the users profile icon (if set)

PR #1017, opened 2025-07-20
add de locale to fastlane

IzzySoft:
Found this in our local tree, but missing here – so sending it up in the hope you find it useful 😉 It's just the German short description, though…


IzzySoft:
Thanks for merging! :hugs:

PR #1018, opened 2025-07-20
compose: home header

jarnedemeulemeester:
Replace the search bar with a header that displays the current server.
Clicking the server allows for switching servers and addresses.

The search bar will still be available from the media page.

### DIFF
--- a/app/phone/src/main/java/dev/jdtech/jellyfin/NavigationRoot.kt
+++ b/app/phone/src/main/java/dev/jdtech/jellyfin/NavigationRoot.kt
@@ -222,9 +222,6 @@ fun NavigationRoot(
             }
             composable<ServersRoute> { backStackEntry ->
                 ServersScreen(
-                    navigateToLogin = {
-                        navController.safeNavigate(LoginRoute())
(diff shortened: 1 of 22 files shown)


Cd16d:
Hi, what about adding a search screen to the navigation bar, with the search bar at the top and, maybe later on, some categories listed below? At the same time, the search bar could be removed from the library screen.

This would make it easier and more intuitive to use the search feature. Also, tapping the search icon in the navigation bar while already on the search screen could automatically open the keyboard, making it quicker to start typing a query.

jarnedemeulemeester:
@Cd16d That's almost exactly what I plan on doing. I was thinking on making the Libraries/Media screen also the search screen. So it's more like a Discover screen.

There would then be 3 main navigation destinations: Home, Discover and Downloads.
I prefer to have exactly 3 destinations so that the Search (or Discover) screen is in the center.

> Also, tapping the search icon in the navigation bar while already on the search screen could automatically open the keyboard, making it quicker to start typing a query.

And yes, that as well. :slightly_smiling_face:

PR #881, opened 2024-11-11
feat: 10.10 Media segments

cd16b:
### Add
- Jellyfin 10.10 MediaSegments support
- Auto Skip options:
  - never
  - only in pip
  - always
- Skip to the next episode if there's no post credit scene
- Show the skip buttons for a specified duration in seconds
- Decide which types of segments should display a skip button or enable auto-skip.
- Basic AndroidTV skip button

### Attention
The old Intro Skipper integration has been removed. The [Intro Skipper](https://github.com/intro-skipper/intro-skipper) plugin supports the new MediaSegmentsApi only on Jellyfin 10.10 or higher.
__________________________________________________
### Issues
Closes #851 
Closes #834 
Fix #833

PR #1019, opened 2025-07-20
add: multi select preference

Cd16d:
Add multi select preference for Mobile and TV.
See #881


jarnedemeulemeester:
@Cd16d I made some small styling changes to the dialog items, as well as use the BaseDialog in more places.
I also added the TV components from the other PR since you seemed to have forgotten them :slightly_smiling_face: 
Lastly I made the Sets of Strings non nullable. I don't think this will break anything.

If there are no objections I think this is ready to merge!

PR #1020, opened 2025-07-24
settings compile errors: remove tint from drawables

cameocoder:
The colorControlNormal attribute could not be resolved and was causing build failures when selecting "Compile all sources" in Android Studio or "gradlew build" from the command line.  The colorControlNormal attribute is resolved in the Android theme, but the settings submodule did not have any theme, so it could not be resolved.

The tint is applied by Jetpack Compose Icon, so it's not needed in the drawable XML anymore.

Note: You could remove all of the settings drawables and replace them with ImageVectors.  However I found it to be difficult to map all of the current drawables to the Material Icons (you may need the extended icons).  If this is desirable, you may need to compromise with a different icon that has the same meaning.

### DIFF
--- a/settings/src/main/res/drawable/ic_closed_caption.xml
+++ b/settings/src/main/res/drawable/ic_closed_caption.xml
@@ -2,8 +2,7 @@
     android:width="24dp"
     android:height="24dp"
     android:viewportWidth="24"
-    android:viewportHeight="24"
-    android:tint="?attr/colorControlNormal">
+    android:viewportHeight="24">
   <path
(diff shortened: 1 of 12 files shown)


jarnedemeulemeester:
Thanks!

I think I will stick with the drawables for now since they seem to work fine without any drawbacks.

It would be nice if there was a library that has all of the Lucide icons as ImageVectors. 
I see there is this one but I have not tested it yet. https://composeicons.com/icon-libraries/lucide
It also does not seem to be updated in a long time.

PR #1021, opened 2025-07-24
Move unused strings from core to their proper submodules.

cameocoder:
This is a continuation of [#1020](https://github.com/jarnedemeulemeester/findroid/pull/1020) trying to get an error-free build.

This commit fixes ExtraTranslation lint errors where there is a string defined for a language that is not defined in the default/English language.

There were a number of strings that existed in core translations that should have been defined in settings or setup submodules.  Most of these were in the Romanian translation.  The settings Romanian translations were also sorted to be in the same order as the English file.

It may be beneficial to do a mass sorting of string files to align them all, but that would result in a large diff and I didn't want to affect any currently open PRs.

With this PR plus #1020 I have a successful build.


jarnedemeulemeester:
Thanks!

I would love to have the strings sorted automatically every time a string is added.
Maybe with a precommit hook or something else. :thinking:

PR #1049, opened 2025-10-05
fix(player): prevent screen from staying on during paused playback

Cd16d:
Adds the FLAG_KEEP_SCREEN_ON window flag when playback starts and removes it when playback is paused. This prevents the screen from staying on while paused.


jarnedemeulemeester:
Thanks a lot!

PR #1057, opened 2025-10-10
fix(player): do not set brightness when exiting PiP mode if V/B gestures are disabled

### LINKED ISSUE #1029: Brightness can't be controlled without gestures
### Describe your issue

I turned off the swipe gestures for brightness and volume control but the brightness of the player does not follow the brightness of the system, nor is there any way to adjust it.
I'd like to use auto brightness and my phone settings instead, but I have to re-enable the gestures just to turn down the brightness.

### Steps to reproduce

1. Turn off gestures
2. Use the player in full screen mode
3. Brightness is locked

### Expected behavior

Brightness should follow the phone's setting when gestures are off (or aren't being manually adjusted yet with gestures on)

### Screenshots

_No response_

### Player

default (ExoPlayer)

### Additional context

_No response_

### Device

Sony Xperia 1 VI (XQ-EC54)

### Android version

15

### App version

v0.15.3

### Jellyfin version

10.10.7

Cd16d:
Do not set brightness when exiting PiP mode if V/B gestures are disabled

May fix #1029

### DIFF
--- a/app/phone/src/main/java/dev/jdtech/jellyfin/PlayerActivity.kt
+++ b/app/phone/src/main/java/dev/jdtech/jellyfin/PlayerActivity.kt
@@ -450,15 +450,17 @@ class PlayerActivity : BasePlayerActivity() {
 
                 // Override auto brightness
                 window.attributes = window.attributes.apply {
-                    screenBrightness =
-                        if (appPreferences.getValue(appPreferences.playerGesturesBrightnessRemember)) {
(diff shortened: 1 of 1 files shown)


jarnedemeulemeester:
Just fixed this in fadcbe39a0fabb0db2220ea856954658aca61616.
I had to add an aditional check in PlayerGestureHelper.
Thanks for bringing this to my attention!

PR #1083, opened 2025-11-30
feat: Add proxy settings support across the application

OpenTools-run:
- Implemented proxy configuration in the settings module, allowing users to enable/disable proxy, set proxy type (HTTP, HTTPS, SOCKS5), and provide authentication details.
- Integrated proxy settings into the network layer by creating a ProxyOkHttpClientFactory to manage OkHttpClient instances with proxy configurations.
- Updated the JellyfinApi to accept a custom OkHttpClient for proxy support.
- Enhanced the AddServerScreen to include a button for navigating to proxy settings.
- Added UI components for proxy settings in both phone and TV applications, including input fields for proxy host, port, username, and password.
- Updated the MPVPlayer to support proxy URLs for network streams.
- Created PreferenceTextInput model for handling text input preferences in settings.
- Added string resources and arrays for proxy settings UI.


OpenTools-run:
it allows to Set a Proxy on the First Run as well before you can even get to "Option"

PR #1089, opened 2025-12-14
Fix NetworkOnMainThreadException in LoginViewModel

saltacc:
Android prohibits network operations on the main thread, which was causing logins to fail. This moves the call to the IO thread, fixing the issue


jarnedemeulemeester:
Thanks for bringing this to my attention.
I have just fixed this in 8d7ba3782226ada7c19a84fa929d96a3567c8fcc

PR #1100, opened 2025-12-24
fix(home): fixed title vertical dimension

Cd16d:
This change adjusts the `ItemCard` component to enforce a single-line title for episodes, even in a vertical layout, to prevent the series name from wrapping and pushing other elements down.

It also introduces a new `@Preview` for the vertical episode card to facilitate easier visualization of this layout.


jarnedemeulemeester:
I actually don't mind the resizing. 
I do agree to limiting the seriesName to a single line in the vertical layout.
But I do not like that the minimum line count for movies in vertical mode is 2. 
It just feels like wasted space to me.

Cd16d:
### New
https://github.com/user-attachments/assets/61e95498-5147-473c-9836-8b70a7392415
### Old
https://github.com/user-attachments/assets/b63f49a2-6cf2-41bc-a453-ce8efb6c14e7

I think having the space always there is better than it appearing and disappearing while side scrolling.
I updated the code to replace minLines with a Column for displaying the text.

jarnedemeulemeester:
I have changed my mind, I do agree that the layout should not shift.

But setting a fixed height with a column does not work for two reasons:
- When you have a movie with a short single line title. The extra space is highlighted when longpressing on a card. This highlight should only include the visible text and not the whitespace.
- Some users may have set a larger system font size. In this case the text may be cut off at the bottom.

Once those are resolved we can move forward to merging this.

However I did just commit some of your changes in 34f7e95584ab6919414d3c66c90e1a4d027ad5e5

Cd16d:
Possible solution: the secondary text (e.g., episode name, year) now occupies a second line, preventing the card from resizing.
- Displays the year for movies.
- Displays the full date range for TV shows.
- Display episode title for episode.
- Makes the primary title bold for better readability (can be removed).
- Makes clickable only the poster (looks cleaner)

### Video
https://github.com/user-attachments/assets/1627f9ef-1880-4af2-883a-04f5f0217456

PR #1101, opened 2025-12-27
fix(episode): remove episode number from season name causing app crash

Cd16d:


jarnedemeulemeester:
We do not want to remove the episode number.
There is just a problem with the strings of some languages.

I'm fixing it at the moment.

PR #1117, opened 2026-01-03
Add missing strings to german + translations

HubGon:
Hi,

with the rework of the app, there are now some missing strings for several languages. I have added those i could find to the German language + translated them directly.

Disclaimer: I used AI-Assistance for searching and adding those strings, but corrected all mistakes it did (double-entries and mistranslations).

Kind regards.


maluueu:
@jarnedemeulemeester can you add these string to [weblate](https://weblate.jdtech.dev/projects/findroid/core/de)?

jarnedemeulemeester:
Thanks!

I do not mind you using AI-assistance as long as you double check and correct the mistakes (which you did).
Thank you for mentioning you did.

I'll add the missing modules to weblate.

PR #1119, opened 2026-01-04
Feat: Option to disable the hold gesture

Rdkang:
## What 
Allows the user to disable the hold gesture by making the setting a select option with:
```
Disabled
2x Speed
Chapter Navigation
```

## Why
This gives the user the choice for more friction to watching in 2x speed to allow them to have better habits.
It also makes the setting easier to understand.

# Image of the changes
EDIT: 2026/07/19 update screenshots to latest changes

Image of the new setting select option
Image of the new setting select values


RyanFoulds:
For the description not showing up in the settings page for your option, this appears to be a deliberate design choice for all settings of this kind. The presentation of those settings is composed in [SettingsSelectCard.kt](https://github.com/jarnedemeulemeester/findroid/blob/main/app/phone/src/main/java/dev/jdtech/jellyfin/presentation/settings/components/SettingsSelectCard.kt) which conspicuously omits writing the description. Compare the composition to `SettingsSwitchCard.kt` in the same package to see the difference.

In my (irrelevant) opinion, since changing that composition to include the description would affect all settings of that type, it should be suggested as a separate change from this PR (if at all).

Cd16d:
I like it.

> Do you think there is a better name for the setting or any of the related wording?

Maybe simply “Hold gesture” would be clearer and more direct.

> Do you think the default (Chapter) for the hold gesture should be a different option?

I think Disabled could work better as the default, to avoid unexpected behavior for new users.

> I can't get the description for the setting option to show up

With `SettingsSelectCard`, the selected option already acts as the description, so in my opinion a separate description isn’t really needed.
If you still want to show extra explanatory text, it could be added inside the selection dialog by changing
`SettingsSelectDialog.kt`, but that would probably require a separate PR, as @RyanFoulds suggested.

Rdkang:
> I like it.
> 
> > Do you think there is a better name for the setting or any of the related wording?
> 
> Maybe simply “Hold gesture” would be clearer and more direct.
> 
> > Do you think the default (Chapter) for the hold gesture should be a different option?
> 
> I think Disabled could work better as the default, to avoid unexpected behavior for new users.
> 
> > I can't get the description for the setting option to show up
> 
> With `SettingsSelectCard`, the selected option already acts as the description, so in my opinion a separate description isn’t really needed. If you still want to show extra explanatory text, it could be added inside the selection dialog by changing `SettingsSelectDialog.kt`, but that would probably require a separate PR, as @RyanFoulds suggested.

Thank you so much @Cd16d and @RyanFoulds for the help, I have now updated the setting title to "Hold gesture", made "Disabled" the default option and removed the description as now the setting is clearer

EDIT: 2026/07/19 update screenshots to latest changes

Image of the new setting select option
Image of the new setting select values

PR #1121, opened 2026-01-06
refactor(nav): improve adaptive navigation

Cd16d:
This commit refactors the adaptive navigation logic to provide a better user experience across different device sizes and orientations.

The new logic introduces the following changes:
- Collapsed navigation rail for landscape phones, PCs, and landscape tablets.
- New navigation bar medium for portrait tablets and foldables.
- Standard bottom navigation bar for portrait phones.
- Vertical centered NavigationItems in landscape

|  | Phone | Tablet |
| -- | -- | -- |
| Portrait | Screenshot_20260113_203210_Findroid Debug | Screenshot_20260106_183933_Findroid Debug |
| Landscape| Screenshot_20260106_183723_Findroid Debug | Screenshot_20260106_183856_Findroid Debug |


jarnedemeulemeester:
Seems like this uses the new flexible navigation bar from material 3 expressive. I am holding off from using any expressive components untill Google has released a stable version of androidx.material3 with those components.
That said it looks like this is part of a stable release...

Also, should we follow the official guidelines of also using a navigation bar in landscape phone view? I do feel like using a navigation bar wastes a lot of vertical screen space. But I just wanted to ask what you think

Cd16d:
Makes sense to wait for a stable material 3 release for the expressive components.

I also completely agree about landscape phones—a bottom nav wastes too much vertical space, so I prefer sticking with the side rail already in the PR.

Also, what if we keep the nav persistently visible on episode and movie screens? Then we could drop the top 'home' button entirely and just use the nav. Let me know what you think!

Issue #1153, opened 2026-02-12
Search no longer shows episodes


Simonxdd:
### Describe your issue

Self-explanatory.

I believe the offending code is `getSearchItems` in `JellyfinRepositoryImpl.kt`, which lacks `BaseItemKind.EPISODE` in the `includeItemTypes`.

Seems like a small oversight from the rewrite.

### Steps to reproduce

1. Search episode name on web.
2. Search episode name in findroid.
3. Web shows result, findroid does not.

### Expected behavior

Episodes should show up, much like in the web interface.

### Device

Google Pixel 8

### Android version

16

### App version

1.0.2

### Jellyfin version

10.11.6

edror12:
@Simonxdd fix suggestion implemented and worked.
Cheers

PR #1158, opened 2026-02-26
Search: include epsiodes

edror12:
Show epsiodes when searching through libraries next to series and movies.

Fixes: #1153


edror12:
Fix tested on physical phone.
Not sure if it's relavant but just in case Pixel 10 running GrapheneOS.

PR #1172, opened 2026-03-17
fix: Load profile picture on phone app.

leo283:
User picture was not showing up on phone app. This PR addresses this by copying the same logic from tv app.


Rdkang:
@leo283 I've been using this pr for the past week, and it's been great thank you :)

PR #1173, opened 2026-03-17
feat(player): implement non-liear brightness sensitivity and decimal display

druneau:
This PR makes it easier to adjust brightness at low levels, which is often difficult on high-brightness screens.

**Non-linear Sensitivity**: Adjustments are now much more precise at low brightness (0-15%), smoothly accelerating to normal speed as brightness increases.

**Decimal Display**: The player UI now shows one decimal place (e.g., `1.2%`) for better visual feedback of these fine adjustments. (1000 steps vs 100)

linear_old
precision_new


Rdkang:
Thank you @druneau this is great, I wouldn't have thought of this and it's fixed a small pain point for me :)

PR #1212, opened 2026-05-23
feat: support jellyfin://item/<id> deep links into the phone app

mooazn:
Hello! 

I'm creating this PR in an effort to support jellyfin deep links on the Findroid phone app.

### Summary

External apps currently cannot deep-link into Findroid. the manifest only declares MAIN/LAUNCHER. This PR adds an ACTION_VIEW filter on MainActivity for `jellyfin://item/<uuid>` so other apps (launchers, automations, the Jellyfin web/desktop ecosystem) can open Findroid directly on a screen. Someone created a PR on the official client to add the same scheme in [jellyfin/jellyfin-android#1933](https://github.com/jellyfin/jellyfin-android/pull/1933). (that has not been merged yet). This PR mirrors that pattern in Findroid.

### Changes

- AndroidManifest.xml: second <intent-filter> on MainActivity. existing MAIN/LAUNCHER untouched. launchMode="singleTask" so re-entry routes through onNewIntent.
- MainActivity.kt: parse intent.data, validate the path as a UUID, resolve via the existing JellyfinRepository.getItem (already DI-provided), and route through the existing navigateToItem helper. onNewIntent covers warm-start. Malformed URIs, unknown ids, and unauthed state all no-op gracefully.
- NavigationRoot.kt: navigateToItem made internal so MainActivity can reuse it (no behavior change for existing call sites). If owners prefer to keep this private, we can wire in a similar logic in MainActivity.

### Notes

- I went with `jellyfin://item/<id>` because Jellyfin item ids are globally unique UUIDs, so the item type can be discovered server-side rather than baked into the URL.
- Right now if no user is signed in, the deep link is dropped rather than stashed across the login flow.

### DIFF
--- a/app/phone/src/main/AndroidManifest.xml
+++ b/app/phone/src/main/AndroidManifest.xml
@@ -34,11 +34,21 @@
         <activity
             android:name=".MainActivity"
             android:exported="true"
+            android:launchMode="singleTask"
             android:windowSoftInputMode="adjustPan">
 
             <intent-filter>
                 <action android:name="android.intent.action.MAIN" />
                 <category android:name="android.intent.category.LAUNCHER" />
+            </intent-filter>
+
+            <intent-filter android:autoVerify="false">
+                <action android:name="android.intent.action.VIEW" />
+                <category android:name="android.intent.category.DEFAULT" />
+                <category android:name="android.intent.category.BROWSABLE" />
+                <data
+                    android:scheme="jellyfin"
+                    android:host="item" />
             </intent-filter>
 
         </activity>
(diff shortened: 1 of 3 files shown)


enter-a-random-username:
I in general I like such an idea, but should be jellyfin standard, but it needs to be written down/implemented at jellyfin first. Reason the PR at jellyfin android got closed.

Maybe open something at Jellyfin or start a discussion, because multiple projects have interest.

e.g. Your current URL will fail with multiple servers + UUID can cause some harm like crash stuff because Image endpoints are unauthenticated and have no limits.

PR #1227, opened 2026-06-17
fix: call use on cursor to ensure it is closed

spacecowboy:
Fixes a bug where the cursor was not closed after use


spacecowboy:
Hi @jarnedemeulemeester  Thanks for the app. Will be great for the summer's road trip. 

I noticed this bug while working on season downloads. Saw a bunch of warnings in logcat.

jarnedemeulemeester (approved):
Looks good, thanks!

PR #1231, opened 2026-06-22
Feat: Watched items toggle button in libary

### LINKED ISSUE #860: Add a hide watched filter

I have lots of movies, so I'd like to be able to hide the ones I've seen already, and have that preference be saved as a default.

I realize that jellyfin has many filters that can be applied to any view, but I think *hide watched* should take precedence over the rest. Even on jellyfin-android-tv, it has its own filter.

Just like sort order, add a hide watched to the dropdown filter, and have that saved in a similar way.

**Describe alternatives you've considered**
NA

**Additional context**
NA

### LINKED ISSUE #1175: Please add a hide watched feature
A clear and concise description of what the problem is. Ex. I'm always frustrated when 

I see a bunch of shows in the library which are no longer relevant to me because I have seen them

A toggle at the top of the libray for show/hide watched

**Describe alternatives you've considered**

**Additional context**

Its such a fundamental feature for an app like this. Please consider implementing.

Rdkang:
Adds a button to the library view to filter out watched items.
The state is stored and is shared between all the libraries like how SortBy is

The Phone UI in action:

[phone ui example](https://github.com/user-attachments/assets/c0eda6a8-0d65-42dd-a83d-e030882ea009)

The TV UI:

[recording of TV ui](https://github.com/user-attachments/assets/abdefe0a-a264-4a17-aeca-2b514a99bbfd)

closes: #860
closes: #1175

### DIFF
--- a/app/phone/src/main/java/dev/jdtech/jellyfin/presentation/film/LibraryScreen.kt
+++ b/app/phone/src/main/java/dev/jdtech/jellyfin/presentation/film/LibraryScreen.kt
@@ -126,6 +126,12 @@ private fun LibraryScreenLayout(
                     }
                 },
                 actions = {
+                    IconButton(onClick = { onAction(LibraryAction.ToggleWatchedFilter(!state.filterWatched)) }) {
+                        Icon(
+                            painter = painterResource(if (state.filterWatched) CoreR.drawable.ic_eye_off else CoreR.drawable.ic_eye),
+                            contentDescription = null,
+                        )
+                    }
                     IconButton(onClick = { showSortByDialog = true }) {
                         Icon(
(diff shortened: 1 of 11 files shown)


midasvo:
I went ahead and let claude yoink it

zeridev:
Would it not be better to have more filter options just like on the jellyfin webui?

It does not have to be all the options. Keep it at the most useful ones at first like the status, filters and release year tabs

Rdkang:
Hi @zeridev, 

The only filter I use is the watched, so I think someone else would be better to develop that as they can test it better :). 

I've implemented the watched filter separately as a quick toggle as it seems to be the most common requested and used filter.

PR #1228, opened 2026-06-18
feat: add ability to download entire season

spacecowboy:
Adds the ability to batch download an entire season. I tried to keep the implementation minimal, reusing as much as possible where it made sense.

* For seasons, both delete and download buttons can be visible at once, because a season might be partially downloaded
* DownloadViewModel has gained the ability to download multiple items. Could also be used to batch download several movies (Movie sets?)
* On download failure, the message is displayed but delete button/download buttons are not rendered. This is existing behavior and I did not change it. If you go out and back in again, then the delete/download buttons are rendered correctly.

Here are some screenshots and such:

## Partially downloaded season renders both Delete and Download buttons
Screenshot From 2026-06-18 09-41-39

## Failure to download an episode in the batch will stop the batch and show message to user
Screenshot From 2026-06-18 09-41-19

## Season is not downloaded, only the download button is rendered
Screenshot From 2026-06-18 09-40-33

## Season is fully downloaded, only the delete button is rendered
Screenshot From 2026-06-18 09-40-27

## User cancels season download

https://github.com/user-attachments/assets/2325bebb-3827-4e3b-b3f2-0e7f00c14a34

## Season download runs out of space, stops and shows error to user

https://github.com/user-attachments/assets/76be1fa5-ac9c-402c-a2b0-df7530193689

Fixes #73
Fixes #461
Fixes #865

PR #1263, opened 2026-08-23
Add an optional sequential download setting

genericJE:
Hi, and thanks for Findroid.

This adds an optional **Sequential** setting for downloads, off by default, on top of the whole season download from #1228 by @spacecowboy.

With the setting on, downloads run one at a time in the order they were queued. A season download then finishes the earliest episodes first rather than splitting the connection across all of them, which on a slow link is the difference between having the first few episodes ready and having ten unfinished ones.

The Sequential setting in the download settings A season downloading, with the first episode running and the rest queued

### What it does

* A Sequential toggle in the download settings, off by default.
* A whole season is submitted as one batch, in episode order.
* The season list shows which episode is downloading and which ones are waiting.
* Downloads keep going while the app is off screen, through a small foreground service.
* The queue is stored, so it survives the process being killed and carries on where it left off.

### Why the queue lives in the app

The system DownloadManager runs everything it is given at once, so the only way to give one item the whole connection is to hold back the next one until the current one reaches a terminal state.

### When the setting is off

`submit()` returns `Bypassed` and a download takes exactly the path it did before, so nothing changes for anyone who does not turn it on.

### On #1228

The season download commit is @spacecowboy's work, kept under his authorship. The loop it introduced is replaced here by the shared queue so that single items and seasons follow the same path. If #1228 lands first I am happy to rebase on it and drop that commit.

### Testing

There are unit tests for the queue. I also ran it on a phone against a deliberately throttled server so the ordering was observable: a season of five episodes queued in episode order, exactly one running with the rest waiting, and each handover happening as the previous episode finished. Killing the app in the middle of a download and reopening it adopted the running download instead of starting a second copy of the same file.

### Transcoded downloads

I have transcoded downloads working on top of this as well, with Jellyfin's own bitrate ladder, a preferred quality setting and quality switching during playback. I left it out here to keep the diff reviewable, but it is on a branch in my fork if it is of interest: https://github.com/genericJE/findroid/tree/feat/transcoded-downloads

Happy to split this up further or change anything you would like done differently.

### DIFF
--- a/app/phone/src/main/AndroidManifest.xml
+++ b/app/phone/src/main/AndroidManifest.xml
@@ -5,6 +5,8 @@
     <uses-permission android:name="android.permission.INTERNET" />
     <uses-permission android:name="android.permission.POST_NOTIFICATIONS" />
     <uses-permission android:name="android.permission.DOWNLOAD_WITHOUT_NOTIFICATION" />
+    <uses-permission android:name="android.permission.FOREGROUND_SERVICE" />
+    <uses-permission android:name="android.permission.FOREGROUND_SERVICE_DATA_SYNC" />
 
     <uses-feature
         android:name="android.hardware.wifi"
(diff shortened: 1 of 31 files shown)


spacecowboy:
Transcoded downloads sounds really useful. There is no need to download a full bluray quality copy to a kid's tablet!

